PR #71, opened 2014-11-12
Gtk backend

mathk:
Hi Heinrich

I have begin to write the Gtk backend. For the time being it is mostly a copy from the WX backend. Let me know if that is ok.

Thanks


ocharles:
@mathk I think this is a great idea, but I don't think it needs to be part of the `reactive-banana` project itself. If I were you, I'd make a new Github repository and take ownership of this - you'll be able to progress it at a much faster pace. I am loosely interested in such a project, so if you are no longer in need of `reactive-banana-gtk` I might be able to help push the project forwards.

mathk:
@ocharles I started this change mostly for fun, if you are interested into it we can start working on a separate project. Feel free to take anything from this PR. Right now I do not have mush time but who knows I might find some to give a help.

ocharles:
Ok, cool. I'm sketching some things out at https://github.com/ocharles/reactive-banana-gtk. Will see where that goes.

Issue #75, opened 2015-02-14
Functor (AnyMoment Event)?


ocharles:
We have this for `AnyMoment Behavior` - is there any reason we can't have it for `Event`?

HeinrichApfelmus:
Should be fine. I think I was simply to lazy to implement it. The `Functor (AnyMoment Behavior)` instance made one particular example significantly nicer, that's why I had included it.

HeinrichApfelmus:
I have added the desired instance.

PR #77, opened 2015-02-22
Adding instance for AMP proposal, allowing GHC 7.10 builds

### LINKED ISSUE #76: reactive-banana doesn't satisfy the applicative-monad proposal
The following is missing:

```haskell
 instance Functor (AnyMoment Identity) where
        fmap = liftM
        
instance Applicative (AnyMoment Identity) where
        pure = return
        (<*>) = ap
```

Apologies for not submitting a pull request - these lines are straight out of my personal NixOS configuration, where I just cat these lines into Switch.hs :)

Twinside:
This should fix #76


ocharles:
This doesn't address #75

Twinside:
My bad, editing comment

Issue #79, opened 2015-03-12
Missing event when combining <@>, filterE, fmap, union, and collect


gelisam:
The following looks complicated, so let me begin by proving that this is a real bug: this is a situation where adding a call to `fmap id` changes the number of events produced by the program!

The reason it is complicated is because it need to be: many different pieces are needed in order to trigger the issue. If I remove the `fmap id`, or the useless looking `filterE (const True)`, or the seemingly unrelated "filtered event" part of the output, the bug disappears.

All right, so here's the code.

    {-# LANGUAGE ScopedTypeVariables #-}
    module Main where

    import Reactive.Banana
    import Reactive.Banana.Frameworks

    network :: forall t. Event t () -> Event t String
    network inputEvent = outputEvent
      where
        appliedEvent :: Event t Int
        appliedEvent = (\_ _ -> 0) <$> lastValue <@> inputEvent
        
        filteredEvent :: Event t Int
        filteredEvent = filterE (const True) appliedEvent
        
        fmappedEvent :: Event t Int
        fmappedEvent = fmap id (filteredEvent)
        
        outputEvent :: Event t String
        outputEvent = (const "filtered event" <$> filteredEvent)
              `union` (show <$> collect (appliedEvent `union` fmappedEvent))
        
        lastValue :: Behavior t Int
        lastValue = stepper 1 $ fmappedEvent

    main :: IO ()
    main = do
        (addHandler, fire) <- newAddHandler
        register (interpretAsHandler network addHandler) putStrLn
        fire ()

So as the single `inputEvent` is fired, an `appliedEvent` with value 0 occurs in response, as does a `filteredEvent` and a `fmappedEvent`, all with the same value. Since all those events are simultaneous and I combine them with `union`, I expect `outputEvent` to occur multiple times, once for each side of the `union`. And indeed, it occurs twice: once with "filtered event", and once with "[0]". But here's the weird thing: if you remove the `fmap id`, or change almost anything else in the code, the second `outputEvent` becomes "[0,0]" instead. Which means it's probably the correct value.

If you replace the `fmap id` with `fmap (const 1)`, the result is still `[0]`, which means that it's the `fmappedEvent` is lost.

gelisam:
[...]

PR #81, opened 2015-04-04
Forget seen elements after they are processed

gelisam:
I am not familiar enough with the internals of reactive-banana to be sure that this is the correct fix for #79, but it does fix that issue and `cabal test` still succeeds.

The idea of the fix is to change `DepsQueue` so that instead of ignoring insertions of elements which have already been inserted at some point in the past, we now only ignore the elements which are currently in the queue. The implementation uses the fact that `insert` is adding the elements it inserts into `seen` and ignores the elements which are already in there; my fix is to remove elements from `seen` once they are dequeued. As a bonus, `seen` is no longer growing without bounds.

I have no idea what this queue is used for, other than the fact that it deduplicates elements and that #79 was introduced when DepsQueue was added. In the context in which it is used, does it make sense to allow the same element to be added to the queue a second time if the element which was first added has already been processed? If you are adding an element which is already in the queue, instead of ignoring it, perhaps it would make sense to readjust its priority?

### DIFF
--- a/reactive-banana/src/Reactive/Banana/Prim/Dependencies.hs
+++ b/reactive-banana/src/Reactive/Banana/Prim/Dependencies.hs
@@ -125,10 +125,10 @@ insert k a q@(DQ queue seen) = {-# SCC insert #-}
         then q
         else DQ (Q.insert k a queue) (Set.insert a seen)
 
-minView :: DepsQueue a -> Maybe (a, DepsQueue a)
+minView :: (Eq a, Hashable a) => DepsQueue a -> Maybe (a, DepsQueue a)
 minView (DQ queue seen) = {-# SCC minView #-} case Q.minView queue of
     Nothing          -> Nothing
-    Just (a, queue2) -> Just (a, DQ queue2 seen)
+    Just (a, queue2) -> Just (a, DQ queue2 (Set.delete a seen))
 
 {-----------------------------------------------------------------------------
     Small tests


gelisam:
Despite the fact that we know that this is not the proper fix for #79, I still think that this change is valid. When `evaluatePulses` traverses the graph, it does so from the smallest to the largest level. Once a particular element at a particular has been traversed, the only elements which can be enqueued afterwards are children of elements at this or higher levels. Since children must have a level which is strictly higher than their parents, all the elements which will be enqueued in the future will have a level which is strictly above the current level. In particular, the current element will never be enqueued again, because its level is too low.

The benefit of merging this change is that the size of `seen` will be the same size as `queue2` instead of growing without bounds.

gelisam:
> The benefit of merging this change is that the size of `seen` will be the same size as `queue2` instead of growing without bounds.

...which only starts to make a difference once the network has many tens of thousands of nodes.

benchmark

Blue is with this pull request, yellow is without. Lower is better. Horizontal axis is number of events in the network (a linear network where each event feeds into the next), vertical axis is the number of seconds for an event to propagate through the entire network.

HeinrichApfelmus:
Ah, ok, I see what you're saying: While is possible to enqueue a child multiple times, justifying the need for `seen`, it is impossible to enqueue a node again after it has been processed. This is correct, because enqueuing a node after it has been processed would mean that there was a parent dependency that we haven't considered — and this would be a bug.

HeinrichApfelmus:
As of version 0.9, this pull request is obsolete. The [function that adds nodes to the queue][1] still tests whether the node has been seen before, but now this test is a lot cheaper, as it simply reads an `IORef`.

Still, I would like to thank you again for tracking down the original bug!

  [1]: https://github.com/HeinrichApfelmus/reactive-banana/blob/master/reactive-banana%2Fsrc%2FReactive%2FBanana%2FPrim%2FEvaluation.hs#L113

PR #83, opened 2015-04-05
add Functor and Applicative for Moment and (AnyMoment Identity)

dnhgff:
this is to satisfy amp for ghc 7.10

i have built this with 7.8.4 as well as 7.10.1


HeinrichApfelmus:
This is the same as #77, right? I have just merged it and will close this one.

(Note that I no longer use the `develop` branch, as people always tend to submit to the `master` branch anyway. :-) I will remove it.)

dnhgff:
ah, i never actually opened up that pr's commits until now but i remember reading the commentary. that commit does not address Moment by itself which is also needed in the test sweet in the second file i change here

PR #84, opened 2015-04-06
amp for Moment in tests

dnhgff:
sorry for the mix up earlyer. this is just the change to get the tests to run under amp as well as the library built


HeinrichApfelmus:
Ah, I didn't see this and made a separate commit just now. But thank you in any case!

PR #86, opened 2015-04-10
remove dead code

gelisam:


gelisam:
As I am reading the code to understand why my other pull request is not the correct fix, I have encountered a few functions which are not used anymore.

HeinrichApfelmus:
The first change sounds good to me, but removing `traverseDependencies` is probably not a good idea — it certainly isn't used anyway, but I consider part of the specification of how the data structure is supposed to work.

gelisam:
Okay, I have removed the second commit from the pull request.

HeinrichApfelmus:
With the release of version 0.9, this particular pull request is now obsolete. But thanks again for your help!

PR #88, opened 2015-04-11
never use ensureAbove

gelisam:
Unlike recalculateParent, ensureAbove wasn't maintaining the invariant
that all children have a higher level than their parent.


gelisam:
Another attempt at fixing #79. This time, I am confident that this is the proper fix.

HeinrichApfelmus:
With the release of version 0.9, this pull request is now obsolete. Thanks again!

PR #112, opened 2015-10-21
added newBehavior function

alios:
Hi! 

I added a ```newBehavior :: a -> MomentIO (Behavior a, Handler a)```  known from sodium.

### DIFF
--- a/reactive-banana/src/Reactive/Banana/Frameworks.hs
+++ b/reactive-banana/src/Reactive/Banana/Frameworks.hs
@@ -30,7 +30,7 @@ module Reactive.Banana.Frameworks (
     -- ** Utility functions
     -- | This section collects a few convience functions
     -- built from the core functions.
-    newEvent, mapEventIO,
+    newEvent, mapEventIO, newBehavior,
 
     -- * Running event networks
     EventNetwork, actuate, pause,
@@ -329,6 +329,14 @@ newEvent = do
     (addHandler, fire) <- liftIO $ newAddHandler
     e <- fromAddHandler addHandler
     return (e,fire)
+
+-- | Build an 'Behavior' toget with an initial value and a 'IO'
+--  action that can push changes to this 'Behavior'.
+newBehavior :: a -> MomentIO (Behavior a, a -> IO ())
+newBehavior s0 = do
+  (e, fe) <- newEvent
+  (,) <$> stepper s0 e <*> pure fe
+
 
 -- | Build a new 'Event' that contains the result
 -- of an IO computation.
--- a/reactive-banana/src/Reactive/Banana/Frameworks.hs
+++ b/reactive-banana/src/Reactive/Banana/Frameworks.hs
@@ -332,7 +332,7 @@ newEvent = do
 
 -- | Build an 'Behavior' toget with an initial value and a 'IO'
 --  action that can push changes to this 'Behavior'.
-newBehavior :: a -> MomentIO (Behavior a, a -> IO ())
+newBehavior :: a -> MomentIO (Behavior a, Handler a)
 newBehavior s0 = do
   (e, fe) <- newEvent
   (,) <$> stepper s0 e <*> pure fe
--- a/reactive-banana/src/Reactive/Banana/Frameworks.hs
+++ b/reactive-banana/src/Reactive/Banana/Frameworks.hs
@@ -330,7 +330,7 @@ newEvent = do
     e <- fromAddHandler addHandler
     return (e,fire)
 
--- | Build an 'Behavior' toget with an initial value and a 'IO'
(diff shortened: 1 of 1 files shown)


jdreaver:
I just started learning reactive-banana after spending a lot of time with Reactive Extensions in Python. I am trying to build GUIs using a model-based approach where the value for each widget is actually stored in a Behavior, and the widget just reflects the changes in the behavior. Then, values can be shared and edited across multiple widgets.

The reason I am commenting is: I have been looking for this exact function! It maps exactly to the Reactive Extensions concepts of `BehaviorSubject(inital_value)` and `BehaviorSubject.on_next(new_value)`. Thanks!

HeinrichApfelmus:
I'm a bit hesitant to add more utility functions to the `Reactive.Banana.Frameworks` module, because ideally, it is only glue code that connects pure FRP to external GUI toolkits. Do you have an example where you need to use it?

@jdreaver I'm not familiar with Reactive Extensions in Python. I would recommend the `stepper` function instead, or is there a reason why it doesn't work in your case?

jdreaver:
I was just trying to create a Behavior from scratch with the ability to push values into it (say, from the callback method of a text widget). I have it all figured out now without this `newBehavior` function, but I looked at the docs and tutorials for hours without being able to figure out that I simply needed `newAddHandler` and an Event derived from that. Obviously this function isn't strictly necessary, but it helped me at that moment :)

alios:
@HeinrichApfelmus I working in reactive-banana with ghcjs for a browser app, which works great as browser/dom is event driven anyways. If I f.e. have a InputField, I want to represent its content as a ```Behavior Text```. DOM lets me register an event which fires when the field gets updated. This is a perfect match for the ```newBehavior`` above. 
I simply create a new Behavior with an initial value (f.e. the empty string), then I register the event listener to the Input field and supply the update function from ```newBehavior``` as Handler.

So this somewhat the counterpart for the polling version which is already there for event driven frameworks.

My other argument would be that sodium provides also both (```newEvent``` and ```newBehavior```) - so adding it would make sense with reactive-banana beeing it's successor (and users so are somewhat asked to migrate).

alios:
@HeinrichApfelmus btw while working on my stuff I've started to introduce a Widget abstraction and create a reactive-banana DOM/HTML binding. As soon as the stuff has stabilized I put that into its own reactive-banana-dom library. 

You can find it here:
https://github.com/open-etcs/openetcs-dmi/blob/master/src/Reactive/Banana/DOM/Widget.hs

But Maybe the Widget abstraction might be a thing to abstract from the DOM context I'am currently using it for. It is primariy based on a WriterMonad which is used to collect cleanup actions during widget tree construction, and executes them (recursivley) when a Widget gets deleted. This is primarily used to deregister event listeners from DOM, but also f.e. to finish Widget related background/worker threads.

An example how this is used would be f.e. here:
https://github.com/open-etcs/openetcs-dmi/blob/master/src/ETCS/DMI/Widgets/Button.hs

This is all still very much work in progress, but as I somewhere in the documentation read, that you also plan to include cleanup handlers to ```Reactive.Banana.Frameworks``` maybe it gives some inseprations.

HeinrichApfelmus:
Hm, I'm still not really sold on this, but I can add it with a haddock comment that includes the implementation. This way, it would have been easier for @jdreaver to figure out which functions to combine.

Also note that in practice, I have found it better to *not* use a Behavior for user input, except in some simple or special cases. For more on this, see my current draft of a ["Widget design guide"][1]. The idea is that the model and visual state of a widget are determined by a Behavior, but that any user input is modeled by an Event, which indicates that the user *suggests* a change, which the model may or may not accept. The whole thing is captured by a data type that I've called [`Tidings`][2].

  [1]: https://github.com/HeinrichApfelmus/threepenny-gui/blob/master/doc/design-widgets.md#three-principles-for-representing-mvc-concepts-with-frp
  [2]: https://github.com/HeinrichApfelmus/threepenny-gui/blob/master/src/Reactive/Threepenny.hs#L342

HeinrichApfelmus:
@alios Ah, the DOM manipulation stuff seems to be very similar to what I'm doing in Threepenny-GUI. If I understand correctly, your `ReactiveDOM` monad corresponds to my `UI` monad. By adding additional information to each DOM element, I was even able to make the function for event registration pure, see [`domEvent`][1].

I haven't looked at cleanup, though, and I haven't gotten around to thinking about widgets very much, yet.

  [1]: http://hackage.haskell.org/package/threepenny-gui-0.6.0.3/docs/Graphics-UI-Threepenny-Core.html#v:domEvent

jdreaver:
Yeah, a comment would probably have been fine.

I really like your idea of `Tidings`, I'm going to check it out!

HeinrichApfelmus:
I've added the function with extensive comment in a separate commit.

PR #117, opened 2015-11-18
alternate interpretFramworks for Behaviors

mlitchard:
I created this to solve my problem of writing tests for Behaviors. Feedback appreciated.

### DIFF
--- a/reactive-banana/src/Reactive/Banana/Combinators.hs
+++ b/reactive-banana/src/Reactive/Banana/Combinators.hs
@@ -39,7 +39,7 @@ module Reactive.Banana.Combinators (
     filterJust, filterApply, whenE, split,
     -- ** Accumulation
     -- $Accumulation.
-    unions, accumB, mapAccum,
+    unions, accumB, mapAccum
     ) where
 
 import Control.Applicative
@@ -308,7 +308,6 @@ split e = (filterJust $ fromLeft <$> e, filterJust $ fromRight <$> e)
     fromLeft  (Right b) = Nothing
     fromRight (Left  a) = Nothing
     fromRight (Right b) = Just b
-
 
 -- $Accumulation.
 -- Note: All accumulation functions are strict in the accumulated value!
--- a/reactive-banana/src/Reactive/Banana/Frameworks.hs
+++ b/reactive-banana/src/Reactive/Banana/Frameworks.hs
@@ -36,7 +36,7 @@ module Reactive.Banana.Frameworks (
     EventNetwork, actuate, pause,
 
     -- * Internal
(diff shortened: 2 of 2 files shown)


mlitchard:
I'm afraid I got a little enthusiastic. I did not mean to have 2 commits in the same merge. I don't knopw how to remove 45b7264 from the merge. I've had second thoughts about ti as it uses partial functions.

mlitchard:
Okay I'm sleepy. Did not commit the problem function I thought I did. TravisCI is not very helpful right now in telling me why the actuall function I submitted failed. In any case I am removing this code and will submit in seperate pull request.

mlitchard:
I'm only trying to merge 42e67a1, don't know how to remove the others

mlitchard:
okay did rebase, all good.

HeinrichApfelmus:
Sorry that it took me so long to have a look at this, I'm currently a little short on time.

1. It's probably not a good idea to do testing in the `MomentIO` monad, because the latter can have arbitrary side effects. It's better to look at `Moment` only, and use the `interpret` function from the `Reactive.Banana.Combinators` module. But apparently, I have made a mistake there: The `interpreter` function does not have the right type signature, I need to fix this.

2. It's possible to test Behaviors by using the function for testing Events.

        interpretE :: (Event a -> Moment (Event b)) -> [a] -> IO [b]
        interpretE = ...

        interpretB :: (Event a -> Moment (Behavior b)) -> [a] -> IO [b]
        interpretB f = interpretE g
            where
            g e = do { b <- f e; return (b <$ e); }

    So, strictly speaking, it's not necessary to include a specialized function. But I can see why it could be useful. Can you give me two or three code examples where you use this function, just so I can see that it's indeed useful?

    That said, the `interpretB` function returns a sligthly differentl result, namely it does not keep track of you "often" the Behavior changes. Now, this information is not supposed to be visible. Do you have an example where you need it, or would the `interpretB` function be enough for your testing purposes?

3. This may be irrelevant if the `interpretB` function is taken instead of the one you wrote, but a more appropriate type for the `interpretFramework'` function is `IO (b,[Maybe b])`. It looks like that is my fault again, because `interpretFramework` already has a suboptimal type, I should change it to `IO [Maybe a]`.

To summarize, there are a couple of things I need to fix. But to better judge which functions are appropriate: it would be very helpful if you could provide me with some example code of your actual tests!

Even though it took me a long time to reply, thank you for bringing this up and dropping me a note about it!

HeinrichApfelmus:
I'm closing this, my changes have found their way into the 1.1.0.0 release. Please feel free to reopen if you feel that I have not addressed everything.

PR #128, opened 2016-02-02
Make MonadFix superclass of MonadMoment

lspitzner:
Since there are exactly two instances, `MomentIO` and `Moment`, both already MonadFix, this should be correct.


lspitzner:
Although i admittedly have not checked if it builds locally, this error is unrelated, is it not?

HeinrichApfelmus:
Sounds like a good idea, thanks! The build error seems to be unrelated to your change.

mitchellwrosen:
Could this be merged?

HeinrichApfelmus:
Thanks!

PR #132, opened 2016-04-07
Add mouse event helpers for reactive-banana-wx

spinda:
These help filter the `mouse` event and expose held `Modifiers` as well as `Point`s.

### DIFF
--- a/reactive-banana-wx/src/Reactive/Banana/WX.hs
+++ b/reactive-banana-wx/src/Reactive/Banana/WX.hs
@@ -16,6 +16,14 @@ module Reactive.Banana.WX (
     -- * Specific widgets
     eventText, behaviorText, eventSelection,
 
+    -- * Mouse events
+    filterModifiers,
+    eventMouseMotion, eventMouseEnter, eventMouseLeave,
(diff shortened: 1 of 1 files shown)


HeinrichApfelmus:
I think that the functions that use `eventMouse` are redundant. I think it's better to export the assorted helper functions `f` under the corresponding name, like

    isMouseMotion :: WX.Event -> Bool

and let the user put things together with `filterE`.

Thinking about it, it seems that `filterJust` may be most appropriate here.

spinda:
Okay. I've changed it so functions like `mouseMotion :: WX.EventMouse -> Maybe (WX.Point, WX.Modifiers)` are exported instead. These can be used with `filterJust` as you suggested:

```haskell
eMouse <- event1 p mouse
let eLeftDrag = filterJust $ leftDrag <$> eMouse
let eShiftDrag = fst <$> second (== justShift) eLeftDrag
```

HeinrichApfelmus:
Thanks! I have changed the documentation slightly and tightening everything into a single commit f1b4df6f7d3fef697793cfae59af08e1549d3c57 .

PR #133, opened 2016-04-07
Add paintB for easy repainting with behaviors in reactive-banana-wx

spinda:
`paintB` is a helper function that updates a widget's `on paint` handler based on a `Behavior` and triggers a `repaint` whenever the `Behavior` changes.

### DIFF
--- a/reactive-banana-wx/src/Reactive/Banana/WX.hs
+++ b/reactive-banana-wx/src/Reactive/Banana/WX.hs
@@ -11,6 +11,7 @@ module Reactive.Banana.WX (
     -- * General
     event1, event0, behavior,
     Prop'(..), sink,
+    paintB,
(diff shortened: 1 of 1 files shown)


HeinrichApfelmus:
The concept seems fine to me, but I would prefer the implementation

    paintB :: Paint w => w -> Behavior (DC () -> Rect -> IO ()) -> MomentIO ()
    paintB w bpaint = do
        sink w [on paint :== bpaint]
        e <- changes bpaint
        reactimate $ repaint w <$ e

as this makes it more clear what is happening (first set the new painting function, then trigger a repaint).

I think I may also prefer a different name: `sinkPaint`. I will make a new commit.

HeinrichApfelmus:
Added. Thanks!

PR #134, opened 2016-04-12
Fix Travis

spinda:
The Travis config is set to test with 1.16, but the minimum supported Cabal version is set to 1.18. This is causing Travis to error out all the time.

Let's try setting the minimum supported Cabal version to 1.16. Will Travis like this better?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ dist/
 *.cabal-sandbox
 *.DS_Store
 *.tmproj
+.stack-work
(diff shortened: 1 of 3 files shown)


spinda:
Well, almost. I guess lspitzner/pqueue#2 needs to be fixed first, as the GHC 8.0.1 build is still failing. Maybe mark this set as an allowed failure until the full release is out?

spinda:
There we go! All good and green again. Let me know if allowing failures for the 8.0.1 tests (for now) is the wrong thing to do here.

HeinrichApfelmus:
Well, if Travis reports that it works with Cabal 1.16, then there is no reason not to lower the bound.

That said, it may actually be that Cabal 1.18 is required for the reactive-banana-wx package. I think that I didn't try to test that one with Travis, though, due to the wx dependencies. Might be worth looking into.

lspitzner:
@HeinrichApfelmus fyi i just released ghc-8.0.1-compatible pqueue-1.3.1.1.

HeinrichApfelmus:
@lspitzner Thanks! Travis has reported that everything works, so I have uploaded a new version of reactive-banana.

PR #137, opened 2016-06-20
LowerUpperSwitch example, demonstrating switchB

DevJac:
This is an example I wrote to help myself understand `switchB`. An example like this would have helped me when I was learning (I'm having some trouble installing wx, so I found the command line examples especially useful).

If you want me to make any improvements before merging, let me know. Perhaps making it interactive like the other command line examples would be worthwhile?

### DIFF
--- a/reactive-banana/doc/examples/LowerUpperSwitch.hs
+++ b/reactive-banana/doc/examples/LowerUpperSwitch.hs
@@ -1,0 +1,36 @@
+{-----------------------------------------------------------------------------
+    reactive-banana
+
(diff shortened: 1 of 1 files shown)


DavidEichmann:
Just a note: you can write:

```
lowerEvent = const toLowerBehavior <$> filterE ("lower" `isPrefixOf`) fireEvent
```
as
```
lowerEvent = toLowerBehavior <$ filterE ("lower" `isPrefixOf`) fireEvent
```
in general
```
const x <$> y == x <$ y
```

PR #154, opened 2017-08-28
Simplify mkWeakIORefValue

DevJac:
This is equivalent to the old code, but without the need to check the version of base.

This change was inspired by looking at the source code for System.Mem.Weak.mkWeak in base.

I will not be offended if you reject this PR. I recognize that there is always risk when changing code, and since this code does the same thing as before, it may not be worth merging.

I have tested this code on GHC 7.8, 7.10, and 8.0. The tests pass on each of those three versions.


mitchellwrosen:
LGTM, thanks. I suppose the original code may have had an eye towards eventually attaching finalizers to some refs, but in the meantime this does clean some things up.

Issue #158, opened 2017-09-17
'coincident' combinator


mitchellwrosen:
I've occasionally found this combinator useful:

```haskell
-- | Only emit event occurrences coincident with the second event.
coincident :: Event a -> Event b -> Event a
coincident e1 e2 = 
  filterRight                  -- (3) Filter down to only coincident events
    (unionWith
      (\(Left x) _ -> Right x) -- (2) Re-tag coincident Lefts as Rights
      (Left <$> e1)            -- (1) Left means "not coincident"
      (unsafeCoerce e2))
 where
  filterRight e = filterJust (either (const Nothing) Just <$> e)
```

I wonder if anyone else has, too?

nickkuk:
What about such simplification?
```haskell
coincident :: Event a -> Event b -> Event a
coincident e1 e2 = filterJust (unionWith const (Nothing <$ e2) (Just <$> e1))
```

mitchellwrosen:
That's the XOR of mine :)

nickkuk:
Ah, sorry, I misunderstood :smile:

nickkuk:
One could suggest more general function
```haskell
intersectionWith :: (a -> b -> c) -> Event a -> Event b -> Event c
```

mitchellwrosen:
Yup, and the implementation is just a variation of `unionWithP`:

```haskell
intersectionWithP :: (a -> b -> c) -> Pulse a -> Pulse b -> Build (Pulse c)
intersectionWithP f px py = do
    p <- newPulse "intersectionWithP" $ eval <$> readPulseP px <*> readPulseP py
    p `dependOn` px
    p `dependOn` py
    return p
  where
    eval (Just x) (Just y) = Just (f x y)
    eval _ _ = Nothing
```

though the name is kind of long... I wish `unionWith` was called `union` instead!

nickkuk:
I think these names are good, because they correspond to similar functions from e.g. Data.Map.

mitchellwrosen:
Well there used to be a `union` function before version 1.0, so the name `unionWith` made more sense. Now that `union` no longer exists, I'd prefer to simply rename `unionWith` to save 4 characters... but that's just a personal preference, and anyways, beside the point of this ticket :)

nickkuk:
[...]

Issue #147, opened 2017-03-13
Name change suggestions


dimitri-xyz:
I think the name of some of the functions and types used in Reactive Banana could be improved to help beginners and also to more closely match the semantic meaning of the underlying operations and types. Here are a few suggestions that I think are important.

Change `actuate` to `activate` — These words have very similar meanings and I believe are interchangeable in the context of `EventNetwork`s. However, activate is much more common. A cursory google search has 246 million hits for activate against only 6.7 million for actuate. So, activate is over 30 times more common. This makes it easier for non-native English speakers to know its meaning.

Change `AddHandler` to `HandlerSet` — I think it is easier to have a type denote nouns (such as a network connection, a list or a hashMap) as opposed to verbs. Those are usually associated with functions (such as in filter, interpret and apply). In the case of `AddHandler`, “Add” is a verb.

Semantically, this types provides a Set of event handlers that are all “fired” together. There is no internal structure, so this is not a list. A HandlerRegistrar or similar name would also work, but the word registrar is not used much. Derived functions such as “newAddHandler” would also have clearer names (e.g. newAddHandler becomes newHandlerSet).

Change `Event` to `EventStream` — I understand this may be a more controversial change. I believe the name `Event` has stuck for historical reasons. However, I believe it is very confusing in the context of Reactive Banana. In common day usage, an event happens only once, whereas in Reactive Banana the type `Event a` encompasses multiple events that happen at different times. From the docs “`Event a` represents a stream of events as they occur in time.” This leads to some confusion. The new name would more closely reflect the type’s semantic meaning. Note that there is no such problem with Behaviors.

I think it is easy to make these changes in the code and documentation, but I wouldn't want to make a pull request before the discussion. To keep backwards compatibility, it is only necessary to make name aliases. For example:

```haskell
type AddHandler = HandlerSet
newAddHandler   = newHandlerSet
fromAddHandler  = fromHandlerSet
actuate         = activate
```

I think Haskell’s Achilles’ heel is its barrier to entry. The intent of this suggestion is to make Reactive Banana easier to learn and understand.

HeinrichApfelmus:
Sorry for the late reply.

Changing `actuate` to `activate` sounds fine to me. (I originally chose the word `actuate` because it has a slightly different connotation that matches better with `pause`.)

[...]

PR #163, opened 2017-09-30
Add 'intersectionWith', 'mergeWith' combinators

mitchellwrosen:
In case you want it, this patch adds `intersectionWith` and `mergeWith` (#158) (though I prefer the names `intersection` and `merge`) Sorry about all the whitespace changes, my editor deletes trailing spaces on save.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -18,7 +18,8 @@ Peter Minten
 Atze van der Ploeg
 Gideon Sireling
 Michael Smith
-Lennart Spitzner	<https://github.com/lspitzner>
+Lennart Spitzner    <https://github.com/lspitzner>
 Henning Thielemann
 Alexey Vagarenko
 Daniel Werner
+Mitchell Rosen      <https://github.com/mitchellwrosen>
--- a/reactive-banana/CHANGELOG.md
+++ b/reactive-banana/CHANGELOG.md
@@ -1,6 +1,12 @@
 Changelog for the `reactive-banana** package
 -------------------------------------------
 
+**unreleased**
+
+* Add `mergeWith` combinator. [#163][]
+
+  [#163] https://github.com/HeinrichApfelmus/reactive-banana/pull/163
+
 **version 1.2.1.0**
 
 * Add `Num`, `Floating`, `Fractional`, and `IsString` instances for `Behavior`. [#34][]
--- a/reactive-banana/src/Reactive/Banana/Combinators.hs
+++ b/reactive-banana/src/Reactive/Banana/Combinators.hs
@@ -19,7 +19,7 @@ module Reactive.Banana.Combinators (
     -- but they are documented at the types 'Event' and 'Behavior'.
     module Control.Applicative,
     module Data.Semigroup,
-    never, unionWith, filterE,
+    never, unionWith, mergeWith, filterE,
     apply,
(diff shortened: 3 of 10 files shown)


HeinrichApfelmus:
Hm, I'm not sold on the `intersectionWith` combinator. If anything, I would prefer `mergeWith` and both `unionWith` and `coincidence` defined in terms of it.

mitchellwrosen:
I've updated this PR to remove `intersectionWith` and re-define `unionWith` in terms of `mergeWith`

mitchellwrosen:
Any thoughts on this @HeinrichApfelmus?

HeinrichApfelmus:
Looks good! Can you add an entry to the changelog and your name to the contributors list (if not already there)? I'll merge if you notify me after that.

mitchellwrosen:
Sure! Can I make one last appeal to shorten the names to `union` and `merge`?

My reasoning is:

- Neither name is imported by `Prelude`.
- In `unionWith`, it's obvious (to me) what the given function does. I'd be similarly unhappy if `filter` was instead called `filterWith`. Of course it filters _with_ a predicate because it's an argument to the function.

  I know that a previous version of `reactive-banana` exported both `union` and `unionWith`, but since `union` is gone, we should reclaim its nice, short name :)
- For `mergeWith`, the "with" is awkwardly referring to three separate arguments.

To avoid a major version bump (if that's a concern) I'd also be totally happy with leaving `unionWith` around, but deprecated in favor of `union`.

Or, if you prefer the names `unionWith` and `mergeWith`, that's fine too :)

HeinrichApfelmus:
The main reasoning behind the names was that they be consistent with the names in the `Data.Map` module. But there may be better arguments in favor of the new names.

Anyway, since name changes require a major version bump and many changes in dependent code, I would like to collect them in a single major release, possibly `2.0`. Could you add the arguments to #147 ?

mitchellwrosen:
Sure, I'll move the discussion. There's also a new map-merge module that I just discovered:

https://hackage.haskell.org/package/containers-0.5.11.0/docs/Data-Map-Merge-Lazy.html

It provides a `merge` function:

    merge    
      :: Ord k	 
      => SimpleWhenMissing k a c	
      -> SimpleWhenMissing k b c	
      -> SimpleWhenMatched k a b c	
      -> Map k a	
      -> Map k b	
      -> Map k c	 

So, name change issue aside, one notable difference is the argument order.

Perhaps `mergeWith` here should be changed to reflect this:

    mergeWith :: (a -> Maybe c) -> (b -> Maybe c) -> (a -> b -> Maybe c) -> Event a -> Event b -> Event c

HeinrichApfelmus:
I like the new order of arguments better.

mitchellwrosen:
Forgot to mention here, I force-pushed with the updated arg order.

mitchellwrosen:
I'm still interested in getting this merged, is there anything else I can do?

ocharles:
If you can add a changelog entry I'm happy to merge this, as it seems to have reached consensus.

mitchellwrosen:
@ocharles Sure thing, done

mitchellwrosen:
I'm going to merge this, there is just that one outstanding rename suggestion I made in the other thread (nix the `*With` suffixes).

Issue #153, opened 2017-08-22
Hook into event/behavior life cycle?


mitchellwrosen:
Hello! Is it possible to detect when an `Event` is no longer relevant to the network (garbage collected)? My use-case is killing a thread that's spawned in a `MomentIO` computation. I can provide more details if necessary. 

Thanks!

HeinrichApfelmus:
This is currently not part of the API, but I think it should be possible to implement. Could you provide a short overview of your use case for why this is a good thing to have? Note that the whole garbage collection / finalizer thing can be rather tricky, so it's always a good idea to avoid it if possible.

mitchellwrosen:
Sure, I was playing around with implementing a sort of arcade, in which some minigames had a concept of a timer and others did not. 

Rather than have a global ticking thread, I wanted to spawn a background thread in the `Moment` that some relevant game is switched to:

```haskell
makeTickEvent :: MomentIO (Event ())
makeTickEvent = do
  (etick, tick) <- newEvent
  
  tid <- liftIO . forkIO . forever $ do
    threadDelay 1000000
    tick ()

  -- Here I want to attach a finalizer to `etick` that kills thread `tid`

  pure etick
```

ocharles:
While this is obviously a fine thing to want to do, can you explain why you'd want to do it without a global tick event? It's easier to add new features if it's clear that they enable functionality that is otherwise impossible.

mitchellwrosen:
It's just the first solution that came to mind, and it'd actually make the code a bit cleaner. 

The real type signature for `makeTickEvent` is

```haskell
makeTickEvent 
  :: Int -- Initial delay
  -> Event (Int -> Int) -- Change delay
  -> MomentIO (Event ())
```

So, the ticks can be sped up or slowed down (in some games) by pressing some keys. If I had to use a global ticking event:

- I'd have an additional behavior to determine whether or not speeding up/slowing down events is possible
- I'd have to reset the delay to the default when switching
- I couldn't guarantee that a tick happens in the first moment of a game's existence

None of these are really game-changers (ha), so a global tick event will actually be fine for my purposes :)

However, now I'm imagining some polling behavior that is truly costly to perform, like periodically hitting some rate-limited API, where only paying for what you use is much more desirable.

mitchellwrosen:
I don't think I need or want this feature anymore, so I'll close the issue.

Issue #162, opened 2017-09-26
`switchE` with an initial event


mitchellwrosen:
Hi, I have some code that I can't quite figure out how to implement with the existing API. I think I want a variant of `switchE` that takes an initial `Event` to act as, until the event-of-events fires.

```haskell
switchE' :: MonadMoment m => Event a -> Event (Event a) -> m (Event a)
```

Semantics are something like this (basically `switchE`, but I think the first event might be special in that if it's emitting in the first moment, it's not dropped. That way, `switchE' e never = pure e`):

```
switchE :: [(Time, a)] -> [(Time, [(Time, a)]) -> Time -> [(Time, a)]
switchE e0 ee t0 =
  case e' of
    []  ->  [(tx, x) | (tx, x) <- e0, t0 <= tx]
    (t1, _) : _ -> [(tx, x) | (tx, x) <- e0, t0 <= tx, tx <= t1] ++ e'
  where
    e' :: [(Time, a)]
    e' = concat [trim t1 t2 e | (t1, t2, e) <- intervals ee, t0 <= t1]

    intervals e :: [(Time, a)] -> [(Time, Time, a)]
    intervals e = [(t1, t2, x) | ((t1, x), (t2, _)) <- zip e (tail e)]

    trim :: Time -> Time -> [(Time, a)] -> [(Time, a)]
    trim t1 t2 e = [(tx, x) | (tx, x) <- e, t1 < tx, tx <= t2]
```

Is this possible to implement somehow?

ocharles:
Are you able to provide a motivating example of why you would need this combinator?

mitchellwrosen:
Sure. Say I have two input events: `Int` (set the factor) and `()` (tick):

```haskell
eFactor :: Event Int <-
  fromAddHandler factorAddHandler

eTick :: Event () <-
  fromAddHandler tickAddHandler
```

Now say my program state is a `Behavior Int` that starts at 0 and increments by the current factor every tick.

```haskell
let eValueGen :: Int -> Moment (Event Int)
    eValueGen fact = do
      val <- valueB bValue
      accumE val ((+ fact) <$ eTick)

-- The initial event: initialize to a factor of 1
eValueFirst :: Event Int <-
  eValueGen 1

-- The event of events to switch to
let eValueNext :: Event (Event Int)
    eValueNext = observeE (eValueGen <$> eFactor)

-- The whole event: behave as eValueFirst until eValueNext fires
eValue :: Event Int <-
[...]

Issue #74, opened 2015-01-15
Consider adding a `once` primitive


ocharles:
We had a discussion about this over on [StackOverflow](https://stackoverflow.com/questions/27816012/performing-a-single-switch-in-reactive-banana/27878929?noredirect=1#comment44315088_27878929).

HeinrichApfelmus:
To recap the problem: The combinator `once` has type and semantics

    once :: Event a -> Moment (Event a)
    once e= \start -> take 1 [(t,a) | (t,a) <- e, start <= t]

It retains the first occurrence when the event starts, but discards the rest.

In principle, it can be implement as follows:

    once e = do
        isFirst <- stepper True $ False <$ e
        return $ whenE isFirst e

However, this implementation is unsatisfactory, because the Event could be *garbage collected* after its first occurrence, because it will be equivalent to `never` later on.

In general, the idea is that Events can be garbage collected if:

1. They have no observable effect.
2. They cannot have occurrences anymore.

The `once` combinator certainly satisfies the second criterion, but the implementation given does not allow for garbage collection, because `filterE` cannot "prove" to the compiler that the event may indeed never fire again.

I cannot find a way to express this in terms of dynamic event switching either. It seems to be necessary to implement this as a new primitive.

HeinrichApfelmus:
This needs significantly more thought than I thought. At the moment, my plan is to release version 1.0 without this combinator and leave the issue open for later.

ocharles:
I just found another bit of code that seems to require `once`. This code builds a command line interface for doing data entry. The idea is we can construct data by using a suitable applicative functor that prompts the user for the individual function arguments. As we provide arguments, we gradually apply the entire function. This can be done by switching between the arguments. Crucially, we only want the switching to go "forward", so we use `once` to ensure once we've provided an argument, we only consider subsequent arguments.

Here's a full example. I need to do a little more work on this to allow for the prompt to run multiple times, but hopefully this helps provide a little more motivation about this combinator.

```haskell
module Main where

import Control.Monad
import Reactive.Banana
import Reactive.Banana.Frameworks
import Text.Read

main :: IO ()
main =
  do (stdinAh,dispatchLine) <- newAddHandler
     compile (do stdin <- fromAddHandler stdinAh
                 runPrompt ((,) <$> name <*> age) stdin >>=
                   reactimate . fmap print) >>=
       actuate
     forever (getLine >>= dispatchLine)

newtype Prompt a =
[...]

PR #165, opened 2017-10-01
Add initial Event to switchE

mitchellwrosen:
This patch changes the type of `switchE` from

```haskell
switchE :: MonadMoment m => Event (Event a) -> m (Event a)
```

to

```haskell
switchE :: MonadMoment m => Event a -> Event (Event a) -> m (Event a)
```

as briefly discussed in https://github.com/HeinrichApfelmus/reactive-banana/issues/162

Consider this a second attempt at a feature request, this time with code attached :)

I'm not fully convinced this change is necessary, as I'm quite new to FRP in general, but this is a pattern that seems to exist in other frameworks. Happy to hear reasons why this is not needed, or if it is, I hope I've implemented it properly! Thanks.


mitchellwrosen:
Is there anything I can do to help move this along?

mitchellwrosen:
@HeinrichApfelmus, any thoughts on this one?

reactive-banana/src/Reactive/Banana/Prim/Combinators.hs
```diff
@@ -120,10 +120,9 @@ executeP p1 b = do
     eval (Just x) = Just <$> liftBuildP (x b)
     eval Nothing  = return Nothing
 
-switchP :: Pulse (Pulse a) -> Build (Pulse a)
-switchP pp = mdo
-    never <- neverP
-    lp    <- stepperL never pp
+switchP :: Pulse a -> Pulse (Pulse a) -> Build (Pulse a)
+switchP p pp = mdo
+    lp <- stepperL p pp
```
mitchellwrosen: Before, the latch-of-pulse had an initial pulse `never`. Now, `p`.

reactive-banana/src/Reactive/Banana/Prim/Combinators.hs
```diff
@@ -134,9 +133,10 @@ switchP pp = mdo
             return Nothing
         -- fetch value from old parent
         eval = readPulseP =<< readLatchP lp
-    
+
     p1 <- newPulse "switchP_in" switch :: Build (Pulse ())
     p1 `dependOn` pp
     p2 <- newPulse "switchP_out" eval
+    p2 `dependOn` p
```
mitchellwrosen: Before, `p2` was initialized with no parents, and would not have a parent until `pp` fired. `p1`, a child of `pp`, would switch `p2`'s parent to the newly fired event.

Now, `p2` does have an initial parent - `p`. This, like before, will get switched away to a new parent by `p1` when `pp` fires.

mitchellwrosen:
In addition to the use case I mentioned in #162, I think I've hit another. In brief:

I have a "game session" that's modeled as:

```haskell
gameSessionMoment :: Event Input -> Moment (Behavior GameState, Event ())
```

where the output `Behavior GameState` is rendered to the screen on changes, and the output `Event ()` fires when the player dies.

I thought about turning this into an infinite game, by simply restarting the whole thing when the player dies. This would mean the final game creation function becomes

```haskell
gameMoment :: Event Input -> Moment (Behavior GameState)
```

(with some out-of-band mechanism for quitting the game, like pressing `Esc`, which is not fed into the FRP network).

I got stuck writing the second function in terms of the first. Here's an attempt:

```haskell
gameMoment :: Event Input -> Moment (Behavior GameState)
gameMoment eInput = do
  (bFirstGameState, eFirstGameOver) <- gameSessionMoment eInput

  -- Event that fires when the current game is over
  let eGameOver = undefined :: Event ()

  let eNewGameSession :: Event (Behavior GameState, Event ())
      eNewGameSession = observeE (gameSessionMoment eInput <$ eGameOver)

  switchB bFirstGameState (fst <$> eNewGameSession)
```

In English:

- Create the first game session.
- When the first game session ends, switch to a new game session.

However I could not figure out how to define `eGameOver` above. If `switchE` took an initial event, then it would be defined as:

```haskell
eGameOver <- switchE eFirstGameOver (snd <$> eNewGameSession)
```

Invisible-Rabbit-Hunter:
Hi, will this ever be merged? I found a case where I think something like this is needed, namely when trying to create a type akin to `Dynamic` in reflex, specifically when creating a `join` operator.

My attempt would look approximately like this
```haskell
data Dynamic a = Dynamic { dynBehavior :: Behavior a, dynEvent :: Event a }
  deriving Functor

joinDyn :: forall m a. MonadMoment m => Dynamic (Dynamic a) -> m (Dynamic a)
joinDyn (Dynamic behDyn evDyn) = do
  d1 <- valueB behDyn
  e <- switchE (dynEvent d1) $ dynEvent <$> evDyn
  b <- switchB (dynBehavior d1) (dynBehavior <$> evDyn)
  return $ Dynamic b e
```
which I think should yield the correct semantics, though I'm actually quite new to FRP in general.

mitchellwrosen:
I'm not sure. Despite having written the patch, I now forget most of what I (thought I) knew about the internals, and it's tough to say without an expert reviewer whether this patch could be considered correct (is there a space leak now? etc).

Perhaps once I have more time I'll continue my side-project of stripping down and refactoring `reactive-banana` (in a separate repo) until the implementation becomes clear, but for now, I can't push this along any further.

ocharles:
To move this forward, I'd like to look at a standalone example of where we really really need this change. It's hard to be sufficiently motivated with where things stand at the moment. This isn't to say "let's not do this", but more "let's make sure we really understand why we're doing this". The game example you provide @mitchellwrosen looks like a great start here, so maybe we should turn that into something we can run and try and find some solutions there. I understand @Invisible-Rabbit-Hunter's suggestion about `Dynamic`, but I'd rather reason in terms of an actual application, rather than an abstraction.

ocharles:
Actually, I'm coming around to this. In https://github.com/HeinrichApfelmus/reactive-banana/issues/162, @HeinrichApfelmus acknowledges that he thinks this is probably something new. In https://github.com/HeinrichApfelmus/reactive-banana/issues/74#issuecomment-530182827, @ChristopherKing42 shows that `once` could be implemented in terms of this. And for me personally I'm yet again building a little UI-toolkit in `reactive-banana`, and wanted a `switchUI :: UI a -> Event (UI a) -> MomentIO (UI a)`. I won't go into the details of what a `UI` is, but suffice to say it's basically Dynamic, which @Invisible-Rabbit-Hunter also expressed a desire for.

So maybe we do have sufficient evidence to move on this. @HeinrichApfelmus, any thoughts?

ocharles:
It may also be worth contrasting other FRP systems. Notably I looked at `reflex` and they have:

```haskell
switchHold :: (Reflex t, MonadHold t m) => Event t a -> Event t (Event t a) -> m (Event t a) 
```

which is precisely the new `switchE` we're talking about here. Interestingly this isn't primitive:

```haskell
switchHold ea0 eea = switch <$> hold ea0 eea
```

where 

```haskell
switch :: Behavior t (Event t a) -> Event t a
hold :: a -> Event t a -> m (Behavior t a)
```

are primitive (at least in the sense they are methods of `Reflex` and `MonadHold`, respectively). `hold` here is our `stepper`, but we don't have a `switch` in this form. So I wonder if that switch is also something to consider. I think with that, our current `switchE` is `switchE e = switch =<< stepper never e`.

ChristopherKing42:
I think the ability to garbage collect old events is a major application. For example, say in a game you have an enemy that can attack the player, with `Event Attack` representing attacks. Once the enemy is gone, this event would never fire again. Without `switchE`, it's impossible (as far as I know) to design it in such a way that this event is garbage collected. With `switchE`, you can switch to `never` when it dies, letting it be garbage collected.

(I do agree though that `switch :: Behavior t (Event t a) -> Event t a` seems even better.)

ChristopherKing42:
Another benefit is to resolve #153.

ChristopherKing42:
`switch :: MonadMoment m => Behavior (Event a) -> m (Event a)` does seem more general then the proposed `switchE`.

ocharles:
@mitchellwrosen I've merged `master` into your branch and also updated the model documentation for `switchE`. It's a bit hairy, maybe there's a better formulation.

reactive-banana/src/Reactive/Banana/Combinators.hs
```diff
@@ -277,12 +277,13 @@ observeE = E . Prim.observeE . Prim.mapE unM . unE
 -- | Dynamically switch between 'Event'.
 -- Semantically,
 --
--- > switchE ee = \time0 -> concat [trim t1 t2 e | (t1,t2,e) <- intervals ee, time0 <= t1]
+-- > switchE e0 ee = \time0 -> [(t, a) | (t,a) <- e0, t < t1] ++ concat [trim t1 t2 e | (t1,t2,e) <- intervals ee, time0 <= t1]
```
ocharles, line 280: This might also be

```suggestion
-- > switchE e0 ee = \time0 -> [(t, a) | (t,a) <- e0, t <= t1] ++ concat [trim t1 t2 e | (t1,t2,e) <- intervals ee, time0 <= t1]
```

We should confirm this before merging.

ocharles (approved):
I'm on board with this, so I'm going this my approval. Tagging in @HeinrichApfelmus for the next review!

HeinrichApfelmus:
(Sorry for the long delay on this.) I think that I'm onboard with the change, especially as @PaulJohnson has given an implementation in terms of existing combinators, https://github.com/HeinrichApfelmus/reactive-banana/issues/162#issuecomment-549797925 , but the new version does have better garbage collection (as `upTo` and `once` are still open).

I main question I have is: Do we want to change the type of `switchE` and break backwards compatibility, or do we want to introduce a new name to keep backwards compatibility for a little longer?

ocharles:
> Do we want to change the type of switchE and break backwards compatibility, or do we want to introduce a new name to keep backwards compatibility for a little longer?

A tricky question! A quick [search on Hackage search](https://hackage-search.serokell.io/?q=switchE%5CW) suggests that there is no Hackage usage of `switchE` - all the results are other libraries defining their own `switchE`. To me the appeal of `reactive-banana` is its compact API, so introducing a new combinator and keeping `switchE` means we are essentially exporting `switchE = switchE2 never`, which seems overly specific. Also, we are the only FRP library that has `switchE` without an initial event, so this change essentially brings us inline with the rest.

If it were me, I'd be inclined to make a breaking change.

reactive-banana/src/Reactive/Banana/Combinators.hs
```diff
@@ -277,12 +277,13 @@ observeE = E . Prim.observeE . Prim.mapE unM . unE
...
 -- >     where
+-- >     t1                 = head [t | (t, _) <- ee]
 -- >     intervals e        = [(time1, time2, x) | ((time1,x),(time2,_)) <- zip e (tail e)]
 -- >     trim time1 time2 e = [x | (timex,x) <- e, time1 < timex, timex <= time2]
```
HeinrichApfelmus, line 284: Neither is correct, I believe, not the least because `t1` is not in scope. 😅  The correct semantics is

```haskell
switchE e0 ee0 time0 =
    concat [ trim t1 t2 e | (t1,t2,e) <- intervals ee ]
  where
    laterThan e time0  = [(timex,x) | (timex,x) <- e, time0 < timex ]
    ee                 = [(time0, e0)] ++ (ee0 `laterThan` time0)
    intervals ee       = [(time1, time2, e) | ((time1,e),(time2,_)) <- zip ee (tail ee)]
    trim time1 time2 e = [x | (timex,x) <- e, time1 < timex, timex <= time2]
```


Formulated in this way, it becomes clearer how `e` aligns with the occurrences in `ee`, and the `trim` function indicates that `<=` instead of `<` is the right choice.

The other main question is: What happens right at `time0`? If the `ee` event has an occurrence at this moment, then this will be ignored — that's why we need the initial event `e` in the first place. Any occurrence of `e` at this moment will also be ignored, however, also by virtue of `trim`.

(That said, the semantics was never quite correct in the case where `ee` is a singleton or the empty list. 🙈 But we can exclude this corner case by proclaiming that the semantics are only supposed to hold in the other cases.)

HeinrichApfelmus (changes requested):
Thanks a lot for the pull request, @mitchellwrosen ! 😊

If somebody could

* Change the comment for `switchE` to use the correct semantics
* Add a major version bump to the `.cabal` and update the changelog

then I think we're ready to merge!

ocharles:
@HeinrichApfelmus Done!

HeinrichApfelmus (approved):

PR #167, opened 2018-02-13
Renaming dfs to reversePostOrder + local var renamings

sgraf812:
I found the function names and surrounding commentary quite unexpected, so I renamed a few things.

### DIFF
--- a/reactive-banana/src/Reactive/Banana/Prim/Dependencies.hs
+++ b/reactive-banana/src/Reactive/Banana/Prim/Dependencies.hs
@@ -1,16 +1,17 @@
 {-----------------------------------------------------------------------------
     reactive-banana
 ------------------------------------------------------------------------------}
-{-# LANGUAGE RecordWildCards, NamedFieldPuns #-}
+{-# LANGUAGE NamedFieldPuns  #-}
+{-# LANGUAGE RecordWildCards #-}
 module Reactive.Banana.Prim.Dependencies (
     -- | Utilities for operating on node dependencies.
     addChild, changeParent, buildDependencies,
     ) where
 
-import Control.Monad
-import Data.Functor
-import Data.Monoid
-import System.Mem.Weak
+import           Control.Monad
+import           Data.Functor
+import           Data.Monoid
+import           System.Mem.Weak
 
 import qualified Reactive.Banana.Prim.Graph as Graph
 import           Reactive.Banana.Prim.Types
@@ -91,7 +92,7 @@ doChangeParent child parent = do
 
     -- lower all parents of the node if the parent was higher than the node
     when (d > 0) $ do
-        parents <- Graph.dfs (P parent) getParents
+        parents <- Graph.reversePostOrder (P parent) getParents
         forM_ parents $ \(P node) -> do
             modify' node $ update levelP (subtract d)
 
--- a/reactive-banana/src/Reactive/Banana/Prim/Graph.hs
+++ b/reactive-banana/src/Reactive/Banana/Prim/Graph.hs
@@ -1,15 +1,15 @@
 {-----------------------------------------------------------------------------
     reactive-banana
-    
+
     Implementation of graph-related functionality
 ------------------------------------------------------------------------------}
 module Reactive.Banana.Prim.Graph where
 
 import           Control.Monad
 import           Data.Functor.Identity
+import           Data.Hashable
 import qualified Data.HashMap.Strict   as Map
 import qualified Data.HashSet          as Set
-import           Data.Hashable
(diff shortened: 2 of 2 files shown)


mitchellwrosen:
FYI, the maintainer can speak for himself, but in my experience these cosmetic PRs to open source projects are not really ever wanted. It puts undue burden on them to verify the absence of any logical changes to the code, which is never easy. Maybe if your patch also came with some tests... ;)

(And again, I do not maintain or speak on behalf of project)

sgraf812:
Well, it's just a refactoring (mostly a renaming even), so I'd expect the test suite to catch any regressions I could have introduced.

mitchellwrosen:
> Well, it's just a refactoring (mostly a renaming even),

That's exactly my point :)

> so I'd expect the test suite to catch any regressions I could have introduced.

Did you verify the code in question is even being tested?

sgraf812:
That chunk of code is quite critical to building the priority queue in which the event framework is to be recomputed, so I think this should be pretty much covered by any end-to-end test.

If anything, I'd rather say that a misnomer like `dfs :: _ -> [a]`, which I'd expect to return some kind of post-order (not its reverse), might induce subtle bugs in past and future call sites. That seems not to be the case for the call site in `*.Dependencies`, just saying.

Actually, I'm OK with this not being merged, just wanted to raise some awareness of the issue.

mitchellwrosen:
I'm sure you are spot on with your analysis and I don't mean to harass you. I'd like for such a patch to get merged, should it prevent confusion due to misnamed things as you suggest, I just feel it's a lot more likely to get merged if either:

1. The diff is _so_ straightforward that it is obviously only renaming variables.
2. It comes with tests.

I happen to suspect the maintainer is busy, as I have two open PRs on this project (with tests!) that have not received any feedback.

mitchellwrosen:
Hello, it's me again, and now I _do_ have commit bits here ;) Uh, reading over the comments I made in the past, I apologize for coming off kind of rude. This time I'll first say thanks for coming along to improve things, and _then_ reiterate that minimal diffs and a clear description of what's happening and why do greatly simplify the reviewer's job ;)

Anywho, I'd like to help push this along if you're still interested @sgraf812. Could you perhaps clarify the intent of this PR once more? Is the name "dfs" simply incorrect here?

(Also - is "reverse post-order" just "pre-order"?)

sgraf812:
No worries :) I agree that it's kind of strange to just drop off code like this, but I figured that a PR would get across the point better than simply opening an issue.

From what I get (I don't really remember opening this PR anymore 😬), it's really just

1. Running `stylish-haskell`, which reorders and formats imports (urgh, feel free to drop)
2. Using `fromMaybe []` instead of `maybe [] id`
3. Rename `dfs` to `reversePostOrder`, because that is what `dfs` computes (in a DFS traversal).
4. Rename a bunch of badly named local vars like `xs` or `ys`

Reverse post-order ist not a pre-order. The [wikipedia page](https://en.wikipedia.org/wiki/Depth-first_search#Vertex_orderings) is helpful on the topic, although it doesn't give an example where rev post-order and pre-order differ (I can provide you an example if you want). Reverse post-order useful to get a topological ordering (disregarding back edges in the DFS resulting from circles in the graph), which in turn is an ordering in which all nodes appear *after* their predecessors. That is the order in whcih you want to update the framework nodes/`Pulse`s.

Anyway, it's not important to me to merge this. I haven't looked at reactive-bananas in 3 years, and I think really just wanted to open "an issue with a point". Feel free to close this if you think it's unnecessary.

mitchellwrosen:
LGTM! @sgraf812  could you please address the merge conflicts if you are still interested in getting this one in? Otherwise, I can take care of it once I'm back at a computer in a couple weeks. 

(Also, if you're in there, I would appreciate reverting the import formatting if that's at all convenient for you. While I'm personally a big fan of formatters, this repo doesn't seem to have a style guide just yet, so I don't think it makes much sense to apply `stylish-haskell` to only one module).

sgraf812:
I'm completely paged out on this and I'm not particularly sure which changes you might want to keep and which you want to discard. So feel free to rip any code you like from this MR.

mitchellwrosen:
No worries then, I'll get this in when I'm back at a computer in s couple weeks. Thanks for the contribution.

HeinrichApfelmus:
Thanks for the contribution! Sorry for being late. 😅 It's indeed difficult to accept such a pull request, especially when I like some names, but prefer other names to stay the same.

I have adopted some changes into commit 6c6424b1d2f1fae172f376204e7df86dc6149796 and merged into master.

PR #168, opened 2018-02-16
ReaderWriterIO: instance Semigroup enables compatibility with GHC-8.4.1

thielema:


HeinrichApfelmus:
Hi Henning! Thanks for the push request! Could you add the `semigroup` package in some way to make it compatible with older GHC versions? Travis CI reports that most builds fail…

thielema:
On Fri, 23 Feb 2018, Heinrich Apfelmus wrote:

> Hi Henning! Thanks for the push request! Could you add the semigroup 
> package in some way to make it compatible with older GHC versions? 
> Travis CI reports that most builds fail…

Sorry, I missed to add 'semigroups' to the Test-Suite. The library part 
already depended on 'semigroups'. Pushed new Cabal file now.

thielema:
Checks passed now!

ocharles:
This looks good to me. @HeinrichApfelmus would you be ok with me cutting a new release including this? We're currently close to getting removed from Stackage as we need to bump our `psqueues` upper bound (which this PR also does).

HeinrichApfelmus:
@ocharles Yes, that would be nice, thank you! You may want to bump the major version too, however, I don't remember what the PVP says about adding superclass constraints.

thielema:
On Wed, 28 Feb 2018, Heinrich Apfelmus wrote:

> @ocharles Yes, that would be nice, thank you! You may want to bump the 
> major version too, however, I don't remember what the PVP says about 
> adding superclass constraints.

Is this related to Semigroup?

Adding superclass constraints would require a major version bump since it 
may break type signatures with too few constraints.

ocharles:
@thielema No, we added `MonadFix` as a super class of `MonadMoment`.

HeinrichApfelmus:
Do I need to click anything? I'm happy for this to go ahead.

ocharles:
I can sort it in the week, but if you want to - go ahead

PR #174, opened 2018-05-13
Add some missing Behavior instances (fixes #34)

### LINKED ISSUE #34: Add Num instance for Behavior
Since GHC 7.4.1 `Eq` and `Show` are no longer superclasses of `Num`.

mitchellwrosen:
Num, Floating, Fractional, IsString

### DIFF
--- a/reactive-banana/CHANGELOG.md
+++ b/reactive-banana/CHANGELOG.md
@@ -1,6 +1,12 @@
 Changelog for the `reactive-banana** package
 -------------------------------------------
 
+**unreleased**
+
+* Add `Num`, `Floating`, `Fractional`, and `IsString` instances for `Behavior`. [#34][]
+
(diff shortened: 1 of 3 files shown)


ocharles:
This is https://github.com/HeinrichApfelmus/reactive-banana/issues/34, and I think we still need to reach an agreement there.

mitchellwrosen:
Indeed, I just decided to throw up a patch in case consensus is reached :)

HeinrichApfelmus:
Hm, yes, ok, let's do this. :smile: Can you add an entry to the changelog? I'm forgot the PVP here, does adding a class instance force a change of the major version?

ocharles:
This is only a minor version bump. As per https://pvp.haskell.org/:

> Non-breaking change. Otherwise, if only new bindings, types, classes, **non-orphan instances** ...

HeinrichApfelmus:
Thanks!

PR #175, opened 2018-05-13
Add more type signatures for readability

mitchellwrosen:
This is just a cosmetic PR, but I feel more type sigs everywhere makes the code easier to understand.


ocharles:
I'm on board with this. @HeinrichApfelmus - good to merge?

HeinrichApfelmus:
Looks good to me. I didn't use type signature, because it was mostly plumbing code, and at the time, I felt it was easier to grasp it without signatures.

PR #182, opened 2018-06-02
RFC: pulse parents are not really weak references

mitchellwrosen:
Hi, I have some code attached to a question I have about the implementation. It seems to me that a pulse child will always keep its pulse parent alive, yet the parents field is a `[Weak SomeNode]`. So, I changed this to a `[SomeNode]` and chased down all of the type errors.

Does this change make sense? Thanks.

### DIFF
--- a/reactive-banana/src/Reactive/Banana/Prim/Dependencies.hs
+++ b/reactive-banana/src/Reactive/Banana/Prim/Dependencies.hs
@@ -45,12 +45,10 @@ buildDependencies (Endo f, parents) = do
 connectChild
     :: Pulse a  -- ^ Parent node whose '_childP' field is to be updated.
     -> SomeNode -- ^ Child node to add.
-    -> IO (Weak SomeNode)
-                -- ^ Weak reference with the child as key and the parent as value.
+    -> IO ()
 connectChild parent child = do
     w <- mkWeakNodeValue child child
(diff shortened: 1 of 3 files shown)


HeinrichApfelmus:
I would have to look this up in my notes again, but the overall idea is that neither should a child keep the parent alive, nor the other way round. The reasoning is:

1. If an external source for events (`AddHandler`) no longer has any event occurrences, then it can be garbage collected. Repeat this process with as many children as possible. (The process ends when `unionWith` was applied.)
2. If an internal event no longer leads to an observable output (this happens when using `switchE` or when not using `reactimate`), then it can be garbage collected. Repeat with as many parents as possible.

I don't quite remember what the current code does exactly, but the idea is too keep both children and parents as weak pointers, for reasons of symmetry. It may be that the current code actually ignores one of the two cases (number 1) because it was easier to implement. I dimly remember that only one of the two cases had to be done explicitly, while the garbage collector could handle the other one, but one had to make an explicit choice.

Sorry for the incomplete answer, this stuff is actually quite tricky.

mitchellwrosen:
Interesting! I think you are correct that case 1 is not currently handled by the code, per the line line I deleted from `connectChild`:

```haskell
mkWeakNodeValue child (P parent)
```

PR #184, opened 2018-06-20
Remove a couple uses of unsafePerformIO

mitchellwrosen:
This patch strips `unsafePerformIO` from the definition of `pureL` and `cachedLatch`. Neither were ever called in a non-IO context so I simply had to translate `let x = ...` to `x <- ...`.


ocharles:
I'm not sure about this. `let x = ...` is very different to `x <- ...` - in particular as to when the `IO` actually happens.

mitchellwrosen:
That's correct - at the time I made the patch I believe I did not see a reason for the extra laziness in these cases, but I can't remember now, and anyways it's probably not worth making this change :)

PR #185, opened 2018-06-24
Add instance Monoid,Semigroup for Behavior

### LINKED ISSUE #183: More missing Behavior instances
```haskell
instance Semigroup a => Semigroup (Behavior a) where
  (<>) = liftA2 (<>)

instance Monoid a => Monoid (Behavior a) where
  mempty = pure mempty
```

mitchellwrosen:
Fixed #183

### DIFF
--- a/reactive-banana/src/Reactive/Banana/Types.hs
+++ b/reactive-banana/src/Reactive/Banana/Types.hs
@@ -94,6 +94,13 @@ instance Applicative Behavior where
 instance Functor Behavior where
     fmap = liftA
 
+instance Semigroup a => Semigroup (Behavior a) where
+  (<>) = liftA2 (<>)
+
+instance (Semigroup a, Monoid a) => Monoid (Behavior a) where
+  mempty = pure mempty
+  mappend = (<>)
+
 instance Num a => Num (Behavior a) where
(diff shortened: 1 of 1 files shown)


ocharles:
As `Behavior` is morally `Time -> a`, we would expect it to have the same monoid instance as that for `a -> b`, which is `Monoid b => a -> b`. I agree with this change. @HeinrichApfelmus OK with you?

reactive-banana/src/Reactive/Banana/Types.hs
```diff
@@ -94,6 +94,13 @@ instance Applicative Behavior where
 instance Functor Behavior where
     fmap = liftA
 
+instance Semigroup a => Semigroup (Behavior a) where
+  (<>) = liftA2 (<>)
+
+instance (Semigroup a, Monoid a) => Monoid (Behavior a) where
```
endgame: This looks odd to me, I would have expected `Monoid a => Monoid (Behavior a)` and `mappend = liftA2 mappend`. is it some backwards-compat idiom from when semigroups was an independent library?

HeinrichApfelmus:
Sorry for the delay. I still exist. 😅 @ocharles Ok with me!

PR #192, opened 2018-10-26
Bring Travis to GHC 8.4.4

gliptak:


mitchellwrosen:
Note that the .travis.yml was auto-generated by https://github.com/haskell-CI/haskell-ci, so the proper way to update it in this way is to adjust `tested-with` in the cabal file and re-run.

(However haskell-ci is a bit fussy and arcane so I think this PR is fine as-is)

gliptak:
@mitchellwrosen thank you for the pointer to **https://github.com/haskell-CI/haskell-ci**

mitchellwrosen:
@gliptak Not from me, I'm not a maintainer of this project

PR #193, opened 2018-12-03
Add MonadUnliftIO instances

hololeap:
These instances would be quite useful for projects also using `UnliftIO`. It would bring in `unliftio-core` as a dependency, but that package only depends on `base` and `transformers`. 

(It does require `base >= 4.5`, which is inconsistent with `reactive-banana`'s dependency. I don't know if this is a problem.)

### DIFF
--- a/reactive-banana/reactive-banana.cabal
+++ b/reactive-banana/reactive-banana.cabal
@@ -50,7 +50,8 @@ Library
                         vault == 0.3.*,
                         unordered-containers >= 0.2.1.0 && < 0.3,
                         hashable >= 1.1 && < 1.3,
-                        pqueue >= 1.0 && < 1.5
+                        pqueue >= 1.0 && < 1.5,
+                        unliftio-core >= 0.1 && < 0.2
 
     exposed-modules:
                         Control.Event.Handler,
@@ -87,6 +88,7 @@ Test-Suite tests
                         HUnit >= 1.2 && < 2,
                         test-framework >= 0.6 && < 0.9,
                         test-framework-hunit >= 0.2 && < 0.4,
+                        unliftio-core >= 0.1 && < 0.2,
                         reactive-banana, vault, containers,
                         semigroups, transformers,
                         unordered-containers, hashable, psqueues, pqueue
--- a/reactive-banana/src/Control/Monad/Trans/RWSIO.hs
+++ b/reactive-banana/src/Control/Monad/Trans/RWSIO.hs
@@ -11,6 +11,7 @@ import Control.Applicative
 import Control.Monad
 import Control.Monad.Fix
 import Control.Monad.IO.Class
+import Control.Monad.IO.Unlift
 import Control.Monad.Trans.Class
 import Data.IORef
 import Data.Monoid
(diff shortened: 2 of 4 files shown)


hololeap:
Another option that someone mentioned on `#haskell` is exposing the constructors for `ReaderWriterIOT`, `MomentIO`, and (possibly) `RWSIOT` in some sort of `.Unsafe` module if taking on the extra dependency is a burden. I just don't see a good way of adding `MonadUnliftIO` instances without having those constructors available.

ocharles:
@HeinrichApfelmus I'm OK with this.

HeinrichApfelmus:
Huh, wait a moment. The class instances appear to work formally, but I think the semantics are questionable. The [unliftio](http://hackage.haskell.org/package/unliftio-0.2.9.0#readme) package specifically notes that

> ### Limitations
>
> Not all monads which can be an instance of MonadIO can be instances of MonadUnliftIO, due to the MonadUnliftIO laws (described in the Haddocks for the typeclass). This prevents instances for a number of classes of transformers:
> * Transformers using continuations (e.g., ContT, ConduitM, Pipe)
> * Transformers with some monadic state (e.g., StateT, WriterT)
> * Transformers with multiple exit points (e.g., ExceptT and its ilk)

The `MonadIO` class definitely falls into the second category: Internally, it will change the state of the event network by adding new nodes, inputs, and outputs.

ocharles:
That's actually fine, it's exactly how state in unlifted moands is meant to work. The point is it has no *monadic* state, which means state threaded by the structure of the monad itself. Note for example that the RIO monad has both MonadState instances and Monadic lifting instances

HeinrichApfelmus:
Sorry, I don't feel comfortable adding `unliftio-core` as a dependency.

The reason is that if `RWSIOT` is a good instance of the `MonadUnliftIO` class, then the following instance is also good:

    instance MonadUnliftIO (StateT s IO) where
        withRunInIO inner = StateT $ \s -> do
          sref <- newIORef s
          a    <- inner $ \m -> do
            s1 <- readIORef sref
            (a, s2) <- runStateT m s1
            writeIORef sref s2
            return a
          readIORef sref

Essentially, the state of the state monad can be repackaged as a "monadic context" (in the language of `unliftio`) by using an `IORef`. The `RWSIOT` monad does use this trick for improving performance.

However, the `unliftio` package does not provide an instance for `StateT` or, more generally, `RWST`; therefore, I do not feel comfortable providing it, either. The `RWSIOT` monad is meant as a more performant substitute for `RWST`, not a semantic enhancement. If a type class does not have an instance for `RWST`, then it should not have an instance for `RWSIOT` either.

That said, I am open to the suggestion of exporting the raw types in an `Unsafe` or `Internal` module, if there is still interest; I do understand the need to dissect the data types of some monads in order to compose them.

----

By the way, according to the laws indicated in the definition of the `MonadUnliftIO` class, the following programs need to be equal, `m1 = m2 = m3`:

    m1, m2, m3 :: RWSIOT () () Char IO ()
    m1 = do
        put 'a'
        s <- (put 'b' >> get)
        liftIO $ print s

    -- use idempotency of `askUnlfitIO`
    m2 = do
        put 'a'
        u <- askUnliftIO
        s <- liftIO $ unliftIO u (put 'b' >> get)
        liftIO $ print s

    -- use that `unliftIO` distributes over `>>`
    m3 = do
        put 'a'
        u <- askUnliftIO
        liftIO $ unliftIO u (put 'b')
        s <- liftIO $ unliftIO u get
        liftIO $ print s

This makes it clear that the `u` has to remember that state put in by the `put 'b'` and make it available to the next `get`. One way to do that is to use an `IORef`.

mitchellwrosen:
Yes, you're absolutely right. I glanced at Moment only long enough to conclude, "yep, it's a ReaderT"... forgetting it's over Build, not IO!

hololeap:
I'm not in need of this anymore, and since it sounds like it wouldn't be a good fit I'm going to close.

PR #200, opened 2019-08-23
Fix build for reactive-banana-wx

gloaming:
It needed a couple of version bumps.

Also this changes to the cabal `custom-setup` stanza which has been around for ages now and IIRC obsoletes the use of `explicit-setup-deps` in stack.yaml.

The bundled `stack.yaml` has been removed; it's too old to be of much use.


HeinrichApfelmus:
Thanks! I take it that everything compiles on your machine and you can run one of the examples?

One issue:

> Also this changes to the cabal custom-setup stanza which has been around for ages now and IIRC obsoletes the use of explicit-setup-deps in stack.yaml

In which version of Cabal was this introduced? We need to constrain the `Cabal-version` field accordingly.

gloaming:
Ah, I was using it for my own demo app, so I didn't think of the bundled examples; they needed some version bumps as well. Fixed that and also the `-Wmissing-home-modules` warning (see [here](https://www.haskell.org/cabal/users-guide/developing-packages.html?highlight=other%20modules#pkg-field-other-modules) and [here](https://www.haskell.org/cabal/users-guide/developing-packages.html?highlight=other%20modules#id2))

The existing `filepath` dependency was a `<=` and I couldn't think why so I changed it to a `<`.

I built and tested all examples with Stack LTS 13.29 and 14.6, everything seemed OK.

```
$ nixos-version
19.09pre191505.7d5375ebf4c (Loris)

$ stack --version
2.1.3.1 x86_64 hpack-0.31.2

$ stack build --flag reactive-banana-wx\:buildexamples 
```
Using the following stack.yaml:
```
packages:
- reactive-banana
- reactive-banana-wx

resolver: lts-14.6

extra-deps:
- cabal-macosx-0.2.4.1
- git: https://github.com/gloaming/wxHaskell
  commit: dd893e8e4456f7bac9260774823050ff4b136119
  subdirs:
  - wx
  - wxc
  - wxcore
  - wxdirect

nix:
  packages: [ mesa.dev wxGTK30 xorg.libX11.dev ]
```

Outside of NixOS, the upstream wxHaskell should work too.

> In which version of Cabal was this introduced? We need to constrain the Cabal-version field accordingly.

Ah, I didn't know about that. Done. It's from 1.24 (release in 2016): https://www.haskell.org/cabal/users-guide/developing-packages.html#custom-setup-scripts

HeinrichApfelmus:
Merging — better late than never! Thank you for your contribution! 😊 

At this time, the [wx][] package seems horribly out-of-date. 🙈 On my machine, `nix-shell -p haskellPackages.wx` fails, because the package `mesa-23.1.9` is marked as broken.

  [wx]: https://hackage.haskell.org/package/wx

PR #205, opened 2020-06-07
Loosen hashable and semigroups' upper bound

poscat0x04:
Right now this package does not build with GHC 8.8.3, please consider revisioning, thanks.


mitchellwrosen:
LGTM, thank you!

PR #207, opened 2021-02-20
Move CI to GitHub Actions

mitchellwrosen:
I noticed CI was failing, so I figured this would be a good opportunity to move over to GitHub Actions.

There are a couple loose ends -

1. I'm just using `ghcup` to install GHC, which is lacking 8.10.4 (but has 8.10.3), and everything prior to 7.10. In the original .travis.yml, we went all the way back to 7.6.

2. The 7.10 build is failing anyway with some "-fPIC"-related errors. I wasn't sure what to do about that, so I just disabled the build for now.

Screen Shot 2021-02-20 at 1 19 22 PM
  
3. I'd like to get 9.0.1 at least attempted in CI, but the build's failing due to pqueue's restrictive base upper bound. We could try relaxing that with e.g. "--allow-newer pqueue:base", but I think it would be nicer not to have to maintain such overrides. Anyways, I just disabled the 9.0.1 build as well because it doesn't seem like GitHub Actions supports the concept of a job that's allowed to fail. You either get a green check or a red X.

Despite all that, I'm decently happy with the CI at the moment, and I think this could be merged. The builds cache both the ~/.cabal/store and dist-newstyle/, so they can be rather fast - the slowest part is spending 20 seconds to generate haddocks (which, incidentally, is not slow on my machine - but I'm not observing the curious output pasted below - so I guess that's another loose end to figure out at some point.)

```
./reactive-banana.cabal has been changed. Re-configuring with most recently
used options. If this fails, please run configure manually.
```


HeinrichApfelmus:
Thank you for updating the CI! Merge at your discretion.

PR #208, opened 2021-04-17
Support building with GHC 9

mitchellwrosen:
This patch puts parens around SCC annotations per the 9.0 migration guide: https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.0#parsing-of-scc-pragmas-less-permissive

Now this codebase builds with GHC 9: `cabal build reactive-banana -w ghc-9.0.1 --allow-newer=pqueue:base`

### DIFF
--- a/reactive-banana/CHANGELOG.md
+++ b/reactive-banana/CHANGELOG.md
@@ -4,8 +4,10 @@ Changelog for the `reactive-banana** package
 **unreleased**
 
 * Add `mergeWith` combinator. [#163][]
+* Make internal SCC pragmas compatible with the GHC 9.0 parser. [#208][]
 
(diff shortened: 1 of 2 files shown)


mitchellwrosen:
All pqueue needed was a base upper bound, I've put up a PR over there: https://github.com/lspitzner/pqueue/pull/40

ocharles:
@mitchellwrosen I think this looks good - shall we get this in?

ocharles (approved):

mitchellwrosen:
@ocharles Sure, I suppose it doesn't hurt to merge even though we still can't actually build with GHC 9 without an override to psqueue's bound :)

... but there's a conflict. Let me take care of that.

PR #209, opened 2021-04-30
Optimize the implementation of Graph.listParents

mitchellwrosen:
This PR optimizes the implementation of `Graph.listParents`. I'll annotate the PR with some comments.

### DIFF
--- a/reactive-banana/src/Reactive/Banana/Prim/Graph.hs
+++ b/reactive-banana/src/Reactive/Banana/Prim/Graph.hs
@@ -67,7 +67,9 @@ listParents gr = list
     where
     -- all nodes without parents
     ancestors :: [a]
-    ancestors = Map.keys (children gr `Map.difference` parents gr)
+    -- We can filter from `children`, because a node without incoming edges can only be in the graph if it has outgoing edges.
+    ancestors    = [x | x <- Map.keys (children gr), not (hasParents x)]
+    hasParents x = Map.member x (parents gr)
     -- all nodes in topological order "parents before children"
     list :: [a]
     list = runIdentity $ dfs' ancestors (Identity . getChildren gr)
--- a/reactive-banana/src/Reactive/Banana/Prim/Graph.hs
+++ b/reactive-banana/src/Reactive/Banana/Prim/Graph.hs
@@ -5,7 +5,14 @@
 ------------------------------------------------------------------------------}
 {-# language ScopedTypeVariables#-}
 
-module Reactive.Banana.Prim.Graph where
+module Reactive.Banana.Prim.Graph
+  ( Graph
+  , emptyGraph
+  , insertEdge
+  , getChildren
+  , listParents
+  , dfs
+  ) where
 
 import           Control.Monad
 import           Data.Functor.Identity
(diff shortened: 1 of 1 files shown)


reactive-banana/src/Reactive/Banana/Prim/Graph.hs
```diff
@@ -5,7 +5,14 @@
 ------------------------------------------------------------------------------}
 {-# language ScopedTypeVariables#-}
 
-module Reactive.Banana.Prim.Graph where
+module Reactive.Banana.Prim.Graph
```
mitchellwrosen, line 8: To make sure some invariants are not violated, made `Graph` an opaque type to the rest of the codebase.

reactive-banana/src/Reactive/Banana/Prim/Graph.hs
```diff
@@ -18,8 +25,19 @@ import           Data.Maybe
     Graphs and topological sorting
 ------------------------------------------------------------------------------}
 data Graph a = Graph
-    { children :: Map.HashMap a [a]
+    { -- | The mapping from each node to the set of nodes reachable by an out-edge. If a node has no out-edges, it is
```
mitchellwrosen, line 28: Just a bit of documentation about how this particular flavor of directed graph representation works.

reactive-banana/src/Reactive/Banana/Prim/Graph.hs
```diff
@@ -47,9 +65,9 @@ getParents gr x = maybe [] id . Map.lookup x . parents $ gr
 listParents :: forall a. (Eq a, Hashable a) => Graph a -> [a]
 listParents gr = list
     where
-    -- all nodes without children
+    -- all nodes without parents
```
mitchellwrosen, line 68: I'm pretty sure this was just a confusing typo, and not a bug in the implementation. We _keep_ nodes that have no parents with this computation:

```haskell
[ x | x <- Set.toList $ nodes gr, null (getParents gr x)]
```

reactive-banana/src/Reactive/Banana/Prim/Graph.hs
```diff
@@ -47,9 +65,9 @@ getParents gr x = maybe [] id . Map.lookup x . parents $ gr
...
     ancestors :: [a]
-    ancestors = [x | x <- Set.toList $ nodes gr, null (getParents gr x)]
+    ancestors = Map.keys (children gr `Map.difference` parents gr)
```
mitchellwrosen, line 70: The minor optimization: 

- Since we only want to identify nodes that have no parents, we can start filtering from `children`, rather than all `nodes`, because a node cannot exist in the graph without at least one outgoing or incoming edge.
- We can simply subtract from all `children` the nodes that have at least one incoming edge, which are the keys of `parents`, since no values of `parents` are an empty list.

mitchellwrosen:
Hmm... uh, CI looks broken. I guess that's my fault due to #207...?

Seems like the line `echo "$HOME/.ghcup/ghc/8.10.3/bin" >> $GITHUB_PATH` isn't working, when running `cabal build` the globally-installed GHC (apparently 9.0) is being picked up.

mitchellwrosen:
A previous build seems to have worked.

https://github.com/HeinrichApfelmus/reactive-banana/runs/2369884533?check_suite_focus=true#step:12:5

I guess something's gone wrong with the caching.

mitchellwrosen:
Heh, well I tried to fix CI in #210, made no progress, and just smuggled in the nuclear option to this PR, which is to take the ~1m hit to install GHC on every run, because caching `~/.ghcup` wasn't working, and I couldn't figure out why.

I'll move that commit over to #210 and merge that for now.

mitchellwrosen:
Ready for review 👍

reactive-banana/src/Reactive/Banana/Prim/Graph.hs
HeinrichApfelmus: Good observation about the `children`!

However, it turns out that using `Data.HashMap.difference` has no special performance benefits — the implementation just builds a new HashMap by looking up the to-be-removed elements. (Difference of `Data.Set.Set` would be a different story). I recommend sticking to the simple list filter.

I have added a comment explaining the optimization, and introduced a new auxiliary function to make the algorithm easier to read:

```suggestion
    -- We can filter from `children`, because a node without incoming edges can only be in the graph if it has outgoing edges.
    ancestors    = [x | x <- Map.keys (children gr), not (hasParents x)]
    hasParents x = Map.member x (parents gr)
```

HeinrichApfelmus:
Looks good, thanks! I have added a suggestion. Not sure what happens if I click the "Commit suggestion" button. 🤔 Could you include it? Then feel free to merge. 😊

PR #210, opened 2021-04-30
attempt to fix CI

mitchellwrosen:


mitchellwrosen:
Hmm, I have no idea what's going on, really.

The `~/.ghcup` cache is only 200 bytes, so clearly it's not caching the installed GHC:

https://github.com/HeinrichApfelmus/reactive-banana/runs/2478552216?check_suite_focus=true#step:4:5

... yet, later on, we see ghc is in fact found in `~/.ghcup`:

https://github.com/HeinrichApfelmus/reactive-banana/runs/2478552216?check_suite_focus=true#step:11:4

How'd it get there, if not from the cache?

And this patch, which is passing CI, is an attempt to fix a different build, which is failing because `ghc-9` / `base-4.15` is being picked up by `cabal build`...

https://github.com/HeinrichApfelmus/reactive-banana/actions/runs/800428920

mitchellwrosen:
Here's a build with even more paranoid printf-debugging: https://github.com/HeinrichApfelmus/reactive-banana/runs/2478685873

I really don't understand why the ~/.ghcup cache that's uploaded is 200 bytes. Maybe it's time to revert to the Travis CI config, unless someone else can come along and figure this one out.

PR #211, opened 2021-04-30
Change 'insertWith (flip (++))' to 'insertWith (++)' in 'insertEdge'

mitchellwrosen:
`(++)` will put the new element at the head of the list, whereas `flip (++)` will put it at the end.

Was there a reason to keep the inserted edges in the order that they were inserted? If so, this patch reverses it. My assumption, based on the fact that it's a directed graph, is the order we store these lists is irrelevant.

### DIFF
--- a/reactive-banana/src/Reactive/Banana/Prim/Graph.hs
+++ b/reactive-banana/src/Reactive/Banana/Prim/Graph.hs
@@ -30,8 +30,8 @@ emptyGraph = Graph Map.empty Map.empty Set.empty
 -- | Insert an edge from the first node to the second node into the graph.
 insertEdge :: (Eq a, Hashable a) => (a,a) -> Graph a -> Graph a
 insertEdge (x,y) gr = gr
-    { children = Map.insertWith (++) x [y] (children gr)
-    , parents  = Map.insertWith (++) y [x] (parents  gr)
+    { children = Map.insertWith (\new old -> new ++ old) x [y] (children gr)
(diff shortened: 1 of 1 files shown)


ocharles:
My guess is this is trying to be like difference list trick and adding to the head was always the intent, but there was confusion about the order of arguments to the map combining function

reactive-banana/src/Reactive/Banana/Prim/Graph.hs
```diff
@@ -30,8 +30,8 @@ emptyGraph = Graph Map.empty Map.empty Set.empty
 -- | Insert an edge from the first node to the second node into the graph.
 insertEdge :: (Eq a, Hashable a) => (a,a) -> Graph a -> Graph a
 insertEdge (x,y) gr = gr
-    { children = Map.insertWith (flip (++)) x [y] (children gr)
-    , parents  = Map.insertWith (flip (++)) y [x] (parents  gr)
+    { children = Map.insertWith (++) x [y] (children gr)
+    , parents  = Map.insertWith (++) y [x] (parents  gr)
```
HeinrichApfelmus, line 34: To avoid confusion in the future, I suggest naming the arguments:

     Map.insertWith (\new old -> new ++ old) x [y]

HeinrichApfelmus:
🤔 I think Charles is right.

reactive-banana/src/Reactive/Banana/Prim/Graph.hs
mitchellwrosen: 👍

PR #212, opened 2021-04-30
Replace foldl with foldl'

mitchellwrosen:


HeinrichApfelmus:
🤔 I must have believed that I found the one valid use of `foldl`. I probably didn't. Thanks!

PR #213, opened 2021-05-28
Cache ghc installation in CI

mitchellwrosen:
Finally got around to figuring out why `~/.ghcup` wasn't caching properly before. Apparently it's a symlink to `/usr/local/ghcup`, and GitHub Actions doesn't follow symlinks when caching.

The fix is just changing `~/.ghcup` to `~/.ghcup/*`


HeinrichApfelmus:
Thanks a lot, Mitchell!

PR #220, opened 2021-09-13
Add `merge`, remove filtering from `mergeWith`

### LINKED ISSUE #203: Suggestion: More general `unionWith`
How about a version of `unionWith` with this more general type:

```unionWith' :: (a -> c) -> (b -> c) -> (a -> b -> c) -> Event a -> Event b -> Event c```

which uses the first function if the first event fires alone, the second function if the second event fires alone, and the third function if they fire simultaneously.

This could theoretically be written using the current `unionWith` using `Either` values, but that would involve incomplete pattern matches, which is kind of ugly. `unionWith` could be written in terms of `unionWith'` rather nicely though. Another nice combinator derived from `unionWith'` is one with the type `(Just a -> Just b -> c) -> Event a -> Event b -> Event c`.

### PR DESCRIPTION
**Pull request by ocharles**
Fixes #203.

### DIFF
--- a/reactive-banana/src/Reactive/Banana/Combinators.hs
+++ b/reactive-banana/src/Reactive/Banana/Combinators.hs
@@ -46,6 +46,7 @@ import Control.Applicative
 import Control.Monad
 import Data.Maybe          (isJust, catMaybes)
 import Data.Semigroup
+import Data.These (These(..), these)
 
 import qualified Reactive.Banana.Internal.Combinators as Prim
 import           Reactive.Banana.Types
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Review by mitchellwrosen — APPROVED** [review]
LGTM. Let's be sure to update Reactive.Banana.Model before releasing, too

**Comment by ocharles** [comment]
Good shout! Model updated.

**Review by mitchellwrosen — APPROVED** [review]

**Comment by mitchellwrosen** [comment]
Oops, model's failing CI

**Comment by ocharles** [comment]
Yep, on it. Don't worry about the messy commit history - I'll squash merge once I'm done :smiley:

File: reactive-banana/src/Reactive/Banana/Combinators.hs [hunk 1]
```diff
@@ -110,20 +111,23 @@ never = E Prim.never
...
-  -> (b -> Maybe c) -- ^ The function called when only the second event emits a value.
-  -> (a -> b -> Maybe c) -- ^ The function called when both events emit values simultaneously.
+  :: (a -> c) -- ^ The function called when only the first event emits a value.
+  -> (b -> c) -- ^ The function called when only the second event emits a value.
+  -> (a -> b -> c) -- ^ The function called when both events emit values simultaneously.
   -> Event a
   -> Event b
   -> Event c
-mergeWith f g h e1 e2 = E $ Prim.mergeWith f g h (unE e1) (unE e2)
+mergeWith f g h e1 e2 = these f g h <$> merge e1 e2
```
vagarenko, line 130: Why not define it in terms of Prim.mergeWith? Just curious.

vagarenko: Then merge can be just
merge = mergeWith This That These

ocharles: Sounds good! I'll make that change

HeinrichApfelmus:
Looks mostly good, though we probably want to export `merge` as well?

In the export list,could you list both `merge` and `mergeWith` somewhere in the "derived combinators" list? The thing is that `merge` can be implemented in terms of `unionWith`:

    merge ea ab = unionWith (\(This a) (That b) → These a b) (This <$> ea) (That <$> eb)

ocharles:
Whoops, yes I meant to export `merge`! I'll add that too. I'm not sure I agree that `merge` is a derived combinator though, as your definition relies on partiality. On the other hand, `unionWith` *can* be derived from `merge` without any partiality. To me `merge` or `mergeWith` is the primitive, and `unionWith` is the derived combinator.

Any more thoughts on where this export should go based on the above?

HeinrichApfelmus:
>  To me `merge` or `mergeWith` is the primitive, and `unionWith` is the derived combinator.

I sympathize and I am tempted by `merge`, but I still would like to label `unionWith` as the primitive version. (It is not necessary to use a partial function as an argument to `unionWith`, the function can be total as the other cases will never occur anyway; the partial version was just easier to put in one line).

The reason is historical: There was a point in time where reactive-banana permitted multiple occurrences per Event, and thus allowed a simple `union :: Event a → Event a → Event a`. In this world, there are many possible semantics of `merge`, and it is not clear which one is better. Labeling `unionWith` as primitive suggests alternative semantics, in this case one where `union` exists.

HeinrichApfelmus (approved):
Looks good to me. Thanks @ocharles ! 😊

ocharles:
That sounds good, thanks for explaining @HeinrichApfelmus !

PR #221, opened 2021-09-14
Change `Event` and `Behavior` to have a representational role

### LINKED ISSUE #219: Add role annotations to `Behavior` and `Event`
Currently these are inferred as `nominal`, but I think the parameter should be representational. This would let us say 

```haskell
newtype CheckboxEvent = CheckboxEvent { toEvent :: Event Bool }
  deriving (Monoid, Semigroup) via Event Any
```

ocharles:
Fixes #219. The main problem is that in `newtype Event a = E (Prim.Event a)`, we have

```haskell
type Prim.Event a = Cached Moment (Pulse a)
data Cached m a = Cached (m a)
```

and that `Cached m a` definition forces `a` to be `nominal`.

This PR works around this nominal role by using a Coyoneda-trick, which restores the role to be representational (at the cost of carrying a mapping function around).

### DIFF
--- a/reactive-banana/src/Reactive/Banana/Combinators.hs
+++ b/reactive-banana/src/Reactive/Banana/Combinators.hs
@@ -78,7 +78,7 @@ which is necessary for the higher-order combinators.
 -- I'm really sorry about the extra 'IO', but it can't be helped.
 -- See source code for the sordid details.
 interpret :: (Event a -> Moment (Event b)) -> [Maybe a] -> IO [Maybe b]
-interpret f xs = Prim.interpret (fmap unE . unM . f . E) xs
+interpret f xs = Prim.interpret (fmap unE . unM . f . mkE) xs
 -- FIXME: I would love to remove the 'IO' from the type signature,
 -- but unfortunately, it is possible that the argument to interpret
 -- returns an Event that was created in the context of an existing network, e.g.
@@ -99,7 +99,7 @@ interpret f xs = Prim.interpret (fmap unE . unM . f . E) xs
 --
 -- > never = []
 never    :: Event a
-never = E Prim.never
+never = mkE Prim.never
 
 -- | Merge two event streams of the same type.
 -- The function argument specifies how event values are to be combined
@@ -123,12 +123,12 @@ mergeWith
   -> Event a
   -> Event b
   -> Event c
-mergeWith f g h e1 e2 = E $ Prim.mergeWith f g h (unE e1) (unE e2)
+mergeWith f g h e1 e2 = mkE $ Prim.mergeWith f g h (unE e1) (unE e2)
 
 -- | Allow all event occurrences that are 'Just' values, discard the rest.
 -- Variant of 'filterE'.
 filterJust :: Event (Maybe a) -> Event a
-filterJust = E . Prim.filterJust . unE
+filterJust = mkE . Prim.filterJust . unE
 
 -- | Allow all events that fulfill the predicate, discard the rest.
 -- Semantically,
@@ -144,7 +144,7 @@ filterE p = filterJust . fmap (\x -> if p x then Just x else Nothing)
 --
 -- This function is generally used in its infix variant '<@>'.
 apply :: Behavior (a -> b) -> Event a -> Event b
-apply bf ex = E $ Prim.applyE (unB bf) (unE ex)
+apply bf ex = mkE $ Prim.applyE (unB bf) (unE ex)
 
 -- | Construct a time-varying function from an initial value and
 -- a stream of new values. The result will be a step function.
@@ -165,7 +165,7 @@ apply bf ex = E $ Prim.applyE (unB bf) (unE ex)
 -- This allows for recursive definitions.
 -- See the discussion below for more on recursion.
 stepper :: MonadMoment m => a -> Event a -> m (Behavior a)
-stepper a = liftMoment . M . fmap B . Prim.stepperB a . unE
+stepper a = liftMoment . M . fmap mkB . Prim.stepperB a . unE
 
 -- | The 'accumE' function accumulates a stream of event values,
 -- similar to a /strict/ left scan, 'scanl''.
@@ -181,7 +181,7 @@ stepper a = liftMoment . M . fmap B . Prim.stepperB a . unE
 -- >     where
 -- >     trimE e start = [(time,x) | (time,x) <- e, start <= time]
 accumE :: MonadMoment m => a -> Event (a -> a) -> m (Event a)
-accumE acc = liftMoment . M . fmap E . Prim.accumE acc . unE
+accumE acc = liftMoment . M . fmap mkE . Prim.accumE acc . unE
 
 {-$recursion
 
@@ -266,7 +266,7 @@ valueBLater = liftMoment . M . Prim.initialBLater . unB
 --
 -- > observeE e = [(time, m time) | (time, m) <- e]
 observeE :: Event (Moment a) -> Event a
-observeE = E . Prim.observeE . Prim.mapE unM . unE
+observeE = mkE . Prim.observeE . Prim.mapE unM . unE
 
 -- | Dynamically switch between 'Event'.
 -- Semantically,
@@ -276,7 +276,7 @@ observeE = E . Prim.observeE . Prim.mapE unM . unE
 -- >     intervals e        = [(time1, time2, x) | ((time1,x),(time2,_)) <- zip e (tail e)]
 -- >     trim time1 time2 e = [x | (timex,x) <- e, time1 < timex, timex <= time2]
 switchE :: MonadMoment m => Event (Event a) -> m (Event a)
-switchE = liftMoment . M . fmap E . Prim.switchE . Prim.mapE (unE) . unE
+switchE = liftMoment . M . fmap mkE . Prim.switchE . Prim.mapE (unE) . unE
 
 -- | Dynamically switch between 'Behavior'.
 -- Semantically,
@@ -284,7 +284,7 @@ switchE = liftMoment . M . fmap E . Prim.switchE . Prim.mapE (unE) . unE
 -- >  switchB b0 eb = \time0 -> \time1 ->
 -- >     last (b0 : [b | (timeb,b) <- eb, time0 <= timeb, timeb < time1]) time1
 switchB :: MonadMoment m => Behavior a -> Event (Behavior a) -> m (Behavior a)
-switchB b = liftMoment . M . fmap B . Prim.switchB (unB b) . Prim.mapE (unB) . unE
+switchB b = liftMoment . M . fmap mkB . Prim.switchB (unB b) . Prim.mapE (unB) . unE
 
 {-----------------------------------------------------------------------------
     Derived Combinators
--- a/reactive-banana/src/Reactive/Banana/Frameworks.hs
+++ b/reactive-banana/src/Reactive/Banana/Frameworks.hs
@@ -168,7 +168,7 @@ reactimate' = MIO . Prim.addReactimate . Prim.mapE unF . unE
 -- this will register a callback function such that
 -- an event will occur whenever the callback function is called.
 fromAddHandler ::AddHandler a -> MomentIO (Event a)
-fromAddHandler = MIO . fmap E . Prim.fromAddHandler
+fromAddHandler = MIO . fmap mkE . Prim.fromAddHandler
 
 -- | Input,
 -- obtain a 'Behavior' by frequently polling mutable data, like the current time.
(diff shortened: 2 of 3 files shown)


Icelandjack:
What if you unfold `Cached` in `Event` + make `Cached` a newtype? Then `Event a` is coercible to `Cached Moment (Pulse a)` and you should have a representable argument.

```haskell
newtype Cached m a = Cached (m a)

newtype Event a = E_ (Moment (Pulse a))
```

Also: Semigroup, Monoid, Num can be derived for `Behaviour` via `Ap Behaviour a` and `Future`/`Moment`/`MomentIO` can newtype derive Functor, Applicative, Monad, MonadFix. 

Screenshot from 2021-09-15 01-41-14

ocharles:
@Icelandjack I need to confirm but I think we might need cached to be `data` for laziness stuff. But if we don't need that, then your idea is great and gives us one less pointer indirection, too!

mitchellwrosen:
Last I looked - and my understanding could be way off here - the `Cached` type is actually not necessary, it just serves as sort of a reminder that some `Cached m` computation (namely the ones built fron `cache`, not `dontCache`) only run the `m` effects at most once. But this of course only requires `MonadIO` and `MonadFix`.

So if I had to speculate wildly, I'd guess `Cached` could be a newtype.

ocharles:
> So if I had to speculate wildly, I'd guess Cached could be a `newtype`.

Let's give it a try then! I'm very keen to chase one fewer pointer if possible. I'll look into this. What @Icelandjack proposes is much nicer than my Yoneda hacking, as cute as it is.

HeinrichApfelmus:
Alas, `Cache` needs to use `data` because we want to memoize the value ("laziness stuff") which is used to implement observable sharing.

Observable allows us to use an `Event` value multiple times without duplicating an `Event` internally. For example, the code

    let e1 = … some event …; e1 = e2 in unionWith const e1 e2

uses the same event internally for `e1` and `e2`; this is not automatic. It would be semantically correct if we duplicate the underlying stream of events of `e1` in `e2`, but less performant.

Observable sharing must use `unsafePerformIO` (or similar), which happens here:

https://github.com/HeinrichApfelmus/reactive-banana/blob/4dcf4b01cdb4cbdcd5347b099a17bda1d025d2e5/reactive-banana/src/Reactive/Banana/Prim/Cached.hs#L33-L36


The fact that `Cache` is `data` ensures that the `unsafePerformIO` is *not* called twice for the same value. Subtle, but very important.

The need for observable sharing likely also rules out other solutions, such as a Yoneda embedding like `E a ~ (something a → r) → r`. The issue is that function arrows do not play well with sharing (though it is not impossible to make it work).

HeinrichApfelmus:
How important is this issue of type roles? If it is a minor improvement, I would recommend against complicating the representation of `Event` or likewise in any way.

If I understand correctly, the issue seems to be that `Cached` used `m a` which is only `nominal` in `a`, because `m` could be a type family. However, we only use the special case `m = Moment`, which is a type synonym where the parameter `a` is representational. If we can override the type role, I recommend we do that, if not, then this example is probably something we should show to the GHC developers and leave it alone for now.

ocharles:
> How important is this issue of type roles? If it is a minor improvement, I would recommend against complicating the representation of Event or likewise in any way.

Not hugely, though I refer back to my original example. This is code I wrote and expected to work, and was a bit disappointed to find I was going to have to hand-write the `Monoid` and `Semigroup` instances.

> If I understand correctly, the issue seems to be that Cached uses m a which is only nominal in a, because m could be a type family. However, we only use the special case m = Moment, which is a type synonym where the parameter a is representational. If we can override the type role, I recommend we do that

Yep, your understanding is correct. I'll see if I can make `m` representational. I wonder also if `data CachedMoment a = CachedMoment (Moment a)` is something that could be considered

mitchellwrosen:
Thanks for the clarification, that is subtle indeed... so subtle that I'm not sure I understand it yet!

Simplifying the machinery a tiny bit, let's say I decorate an IO action with an `unsafePerformIO` preamble, just like `cache`:

```haskell
{-# NOINLINE cache #-}
cache :: IO a -> IO a
cache action =
  unsafePerformIO do
    putStrLn "hello, world!"
    pure action
```

I should still expect this function to return a thunk that, when forced to WHNF, prints "hello, world" and returns the IO action passed in. So regardless of how many pointers to the returned action exist, the thunk will only be updated in place once.

Sorry for dragging this on, feel free to ignore this comment

HeinrichApfelmus:
> I should still expect this function to return a thunk that, when forced to WHNF, prints "hello, world" and returns the IO action passed in. So regardless of how many pointers to the returned action exist, the thunk will only be updated in place once.

If values of type `IO a` were thunks, this would be correct. But `IO` gets special treatment from the compiler. I just checked that it works as you expect it to work in the interpreter, but I don't think that it will (always) work in compiled code.

I remember recording a remark on this in the [HeinrichApfelmus/optimize-monad-trans](https://github.com/HeinrichApfelmus/optimize-monad-trans) repository at one point:

> Note that the same reasoning could be applied to the state token for the IO monad. Why doesn't GHC build closures for that? The answer is that the compiler treats the State# type constructor as a special case and doesn't refrain from recomputing IO actions. This is GHC's infamous [state hack](https://gitlab.haskell.org/ghc/ghc/-/issues/1168).

HeinrichApfelmus:
In the [documentation on role annotations](http://downloads.haskell.org/~ghc/latest/docs/html/users_guide/exts/roles.html), a data type that is identical to `Cache m a` is mentioned specifically. They call it `Tricky a b`. :joy: I don't suppose that there is a way to tell the compiler that `m` can never be a type synonym? There could be a StackOverflow question for that. The case where `m` is a monad parameter actually seems like a fairly common use case to me.

tstat:
Does the "state hack" thwart the sharing attempt in this case? I don't see how yet:

As I understand it, the "state hack" is to label all lambdas binding `State#` as one shot so that a let on the outside will be floated into the lambda, thus eta-expanding the state lambda beyond the let. This comes back to bite us when the we execute that IO action multiple times. A small example would be:

```haskell
expensive :: Int -> Int
expensive n = n
{-# NOINLINE expensive #-}

stateHack :: Int -> State# RealWorld -> (# State# RealWorld, Int #)
stateHack n =
  let i = expensive n
   in (\s -> (# s, i #))
```

where `stateHack` produces the core:

```haskell
stateHack
  = \ (n_avF :: Int) (s_avH :: State# RealWorld) ->
      (# s_avH, expensive n_avF #)
```
and we see that our state lambda was eta-expanded outside of our let. 

Compare this with a non-state argument:

```haskell
normal :: Int -> () -> (# (), Int #)
normal n =
  let i = expensive n
   in (\s -> (# s, i #))
```

and we see that `expensive` is shared across calls of the returned function, as expected

```haskell
normal
  = \ (n_avI :: Int) ->
      let {
        i_sFV :: Int
        [LclId]
        i_sFV = expensive n_avI } in
      \ (s_avK :: ()) -> (# s_avK, i_sFV #)
```

We aren't discussing the usual "state hack" problem of a let expression returning an `IO` function. We are using `unsafePerformIO` which itself uses `runRW#`, and `runRW#` is not inlined until after all core-to-core passes. 

So, let's instead consider a variant of `expensive` and `stateHack` where `expensive` is in `IO` and we run it with `runRW#`:

```haskell
expensiveIO :: Int -> State# RealWorld -> (# State# RealWorld, Int #)
expensiveIO n s = (# s, n #)
{-# NOINLINE expensiveIO #-}

stateHackIO :: Int -> (State# RealWorld -> (# State# RealWorld, Int #))
stateHackIO n =
  -- newAction :: IO (IO Int)
  let newAction :: State# RealWorld -> (# State# RealWorld, State# RealWorld -> (# State# RealWorld, Int #) #)
      newAction st0 = case expensiveIO n st0 of
        (# st1, i #) -> (# st1, \s -> (# s, i #) #)
   in case runRW# newAction of
        (# _, a #) -> a
```
In this variant of `stateHack` the returned state lambda is _inside a  tuple_, and `expensiveIO n st0` cannot be inlined into the state lambda because the resulting state token, `st1`, is used in the unboxed tuple that contains the state lambda.

Peeking at the core reveals that we get the desired sharing:

<details>
<summary>Core</summary>

```haskell
stateHackIO
  = \ (n_axw :: Int) ->
      case runRW#
             @ ('TupleRep '[ 'TupleRep '[], 'LiftedRep])
             @ (# State# RealWorld,
                  State# RealWorld -> (# State# RealWorld, Int #) #)
             (\ (st0_axy [OS=OneShot] :: State# RealWorld) ->
                case expensiveIO n_axw st0_axy of { (# ipv_syX, ipv1_syY #) -> -- execute `expensiveIO` and bind result to `ipv1_syY`
                (# ipv_syX,
                   \ (s_axB :: State# RealWorld) -> (# s_axB, ipv1_syY #) #) -- `ipv1_syY` is captured in the returned IO function
                })
      of
      { (# ipv_sz1, ipv1_sz2 #) ->
      ipv1_sz2
      }
```

</details>

Now consider a cache example using `IORef`

```haskell
cache :: IO a -> IO a
cache action =
  unsafePerformIO do
    ref <- newIORef Nothing
    let newAction = do
          readIORef ref >>= \case
            Nothing -> do
              res <- action
              writeIORef ref (Just res)
              pure res
            Just res -> pure res
    pure newAction
```

We are returning an `IO` function, that is, a `State#` bound lambda, and it seems to me that the important question would be: "Does the returned `State#` lambda eta-expand past the creation of the `IORef`?" I don't see how it would, and the core I compiled below shows that it did not. I think this is because of the reasons I outlined in the simpler example above, namely, that the returned `IO` function is inside an unboxed tuple that provides no eta-expansion opportunity, and `runRW#` is not inlined until `CorePrep`.

<details>
<summary>Core</summary>

```haskell
cache
  = \ (@ a_axD) (action_awc :: IO a_axD) ->
      case runRW#
             @ ('TupleRep '[ 'TupleRep '[], 'LiftedRep])
             @ (# State# RealWorld, IO a_axD #)
             (\ (s_aE8 [OS=OneShot] :: State# RealWorld) ->
                case noDuplicate# @ RealWorld s_aE8 of s'_aE9 { __DEFAULT ->
                case newMutVar#
                       @ (Maybe a_axD) @ RealWorld (GHC.Maybe.Nothing @ a_axD) s'_aE9
                of
                { (# ipv_aET, ipv1_aEU #) ->  -- Our IORef is created here
                (# ipv_aET,
                   (\ (s1_aEK :: State# RealWorld) -> -- And here is our State# lambda capturing the IORef
                      case readMutVar# @ RealWorld @ (Maybe a_axD) ipv1_aEU s1_aEK of
                      { (# ipv2_XF8, ipv3_XFa #) ->
                      case ipv3_XFa of {
                        Nothing ->
                          case (action_awc
                                `cast` (GHC.Types.N:IO[0] <a_axD>_R
                                        :: IO a_axD
                                           ~R# (State# RealWorld -> (# State# RealWorld, a_axD #))))
                                 ipv2_XF8
                          of
                          { (# ipv4_XFi, ipv5_XFk #) ->
                          case writeMutVar#
                                 @ RealWorld
                                 @ (Maybe a_axD)
                                 ipv1_aEU
                                 (GHC.Maybe.Just @ a_axD ipv5_XFk)
                                 ipv4_XFi
                          of s2#_aGj
                          { __DEFAULT ->
                          (# s2#_aGj, ipv5_XFk #)
                          }
                          };
                        Just res_axr -> (# ipv2_XF8, res_axr #)
                      }
                      })
                   `cast` (Sym (GHC.Types.N:IO[0] <a_axD>_R)
                           :: (State# RealWorld -> (# State# RealWorld, a_axD #))
                              ~R# IO a_axD) #)
                }
                })
      of
      { (# ipv_aEb, ipv1_aEc #) ->
      ipv1_aEc
      }
```

</details>

This can perhaps be seen even more plainly in the STG:

<details>
<summary> STG </summary>

```
    \r [action_sIE]
        case
            case noDuplicate# [GHC.Prim.realWorld#] of s'_sIF [Occ=Once] {
            __DEFAULT ->
            case newMutVar# [GHC.Maybe.Nothing s'_sIF] of {
            (#,#) ipv_sIH [Occ=Once] ipv1_sII ->  -- We create our IORef and bind it to ipv1_sII here
            let {
              sat_sIU [Occ=Once] :: GHC.Types.IO a_axD -- We allocate a closure for our returned IO function here, capturing our IORef
              [LclId] =
                  \r [s1_sIJ] -- Here is the state lambda
                      case readMutVar# [ipv1_sII s1_sIJ] of {
                      (#,#) ipv2_sIL [Occ=Once*] ipv3_sIM [Occ=Once!] ->
                      case ipv3_sIM of {
                        GHC.Maybe.Nothing ->
                            case action_sIE ipv2_sIL of {
                            (#,#) ipv4_sIP [Occ=Once] ipv5_sIQ ->
                            let {
                              sat_sIR [Occ=Once] :: GHC.Maybe.Maybe a_axD
                              [LclId] =
                                  CCCS GHC.Maybe.Just! [ipv5_sIQ];
                            } in
                              case
                                  writeMutVar# [ipv1_sII sat_sIR ipv4_sIP]
                              of
                              s2#_sIS [Occ=Once]
                              {
                              __DEFAULT -> (#,#) [s2#_sIS ipv5_sIQ];
                              };
                            };
                        GHC.Maybe.Just res_sIT [Occ=Once] -> (#,#) [ipv2_sIL res_sIT];
                      };
                      };
            } in  (#,#) [ipv_sIH sat_sIU];
            };
            }
        of
        {
        (#,#) _ [Occ=Dead] ipv1_sIX [Occ=Once] -> ipv1_sIX;
        };
```

</details>

PR #223, opened 2021-09-16
Add `Semigroup` and `Monoid` instances to `Moment` and `MomentIO`

### LINKED ISSUE #222: Add `Monoid a => Monoid (MomentIO a)` (and `Moment`)
This instance would sequence `MomentIO` computations and combine their results under `Monoid a`. This is useful as it allows us to write things like:

```haskell
e <- mconcat
  [ switchE ...
  , switchE ..
  , pure ...
  ]
```

This produces a single `Event` that's the combination of two events that come from `switchE`, and one other event. All these events are combined under the `Monoid a => Monoid (Event a)` instance.

The current alternative is to write:

```haskell
e <- mconcat <$> sequence
  [ switchE ...
  , switchE ..
  , pure ...
  ]
```

which is not awful, but more verbose than it needs to be.

This instance is motivated from wanting it to exist while writing some `reactive-banana` code right now, but also the discussion by Gabriella (@Gabriel439) [about these instances on Twitter](https://twitter.com/GabriellaG439/status/1436716446881517571).

ocharles:
Fixes #222

### DIFF
--- a/reactive-banana/CHANGELOG.md
+++ b/reactive-banana/CHANGELOG.md
@@ -5,6 +5,8 @@ Changelog for the `reactive-banana** package
 
 * Added `Semigroup` and `Monoid` instances to `Moment` and `MomentIO`. [#223][]
 
+  [#223]: https://github.com/HeinrichApfelmus/reactive-banana/pull/223
+
 **Version 1.2.2.0**
 
 * Optimize the implementation of `Graph.listParents` [#209][]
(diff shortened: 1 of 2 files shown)


mitchellwrosen (approved):
LGTM; perhaps it could use a changelog entry too?

ocharles:
Indeed it could, I'll add one now

HeinrichApfelmus (approved):
The main issue with adding a `Monoid` instance is that there are often many competing ways in which a type can be turned into a monoid. After all, the mathematical definition of a monoid requires the specification of both a set of elements and an operation; specifying the set alone is not enough.

If we add an instance, it has to be "canonical", i.e. a very natural default choice. Fortunately, I think that this instance is a good default choice, specifically because it only makes a judgement about the monad `m`, and defers judgment about the type `a`.

I was thinking that for any monad `m`, the type `m ()` is a monoid via `(>>)`. But this case is actually subsumed by the virtue of `()` being a monoid!

Also the particular monad considered here is not an instance of `Alternative`, which would correspond to another canonical monoid structure.

ocharles:
@HeinrichApfelmus all great points and I agree with all of them ☺️ It sounds like here we're safe though, as you say (there doesn't appear to be any other _obvious_/canonical choice)

PR #224, opened 2021-09-16
Make the `Eq` instance for `SomeNode` total

### LINKED ISSUE #204: `instance Eq SomeNode` missing case coverage
In `Reactive.Banana.Prim.Types` we have the following code:

`instance Eq SomeNode where`
`    (P x) == (P y) = equalRef x y`
`    (L x) == (L y) = equalRef x y`
`    (O x) == (O y) = equalRef x y`

This is missing coverage of the other cases. It should have this added:
`   _  == _  = False`

ocharles:
Fixes #204.

### DIFF
--- a/reactive-banana/src/Reactive/Banana/Prim/Types.hs
+++ b/reactive-banana/src/Reactive/Banana/Prim/Types.hs
@@ -139,6 +139,7 @@ instance Eq SomeNode where
     (P x) == (P y) = equalRef x y
     (L x) == (L y) = equalRef x y
     (O x) == (O y) = equalRef x y
+    _     == _     = False
 
 {-# INLINE mkWeakNodeValue #-}
 mkWeakNodeValue :: SomeNode -> v -> IO (Weak v)


mitchellwrosen (approved):
Seems fine, though the code could be considered safer as-is, if unlike SomeNodes are never meant to be compared for equality because the caller should somehow know, at every such comparison what the underlying constructor actually is

ocharles:
That's a great point. If it's true that we'd rather crash I'll amend this PR to have a comment being explicit, which I think will still let us close #204.

HeinrichApfelmus:
Nodes with different constructors are indeed never meant to be compared, but that may not be obvious from the code. The main reason is that all the `SomeNode` go into a `Graph` structure for tracking dependencies, which needs `Eq`. I think that `O` and `L` do go into the structure, but they are all leaves. To find out what whether this is true, one has to look at the calls to `addChild`.

Anyway, I think that adding the extra case is semantically correct and therefore cannot hurt. Since the `Eq` is not irrelevant for the `HashMap`, I think this may actually be a good idea. 🤔 Not sure why the library never crashed before.

HeinrichApfelmus (approved):

ocharles:
We thought this was never possible but...

```
ExceptionInLinkedThread (ThreadId 128) src/Reactive/Banana/Prim/Types.hs:(139,5)-(141,33): Non-exhaustive pat
terns in function ==
```

That's exactly the function fixed in this PR :scream: There might be a bug somewhere else, and I can only cause this to happen if I spawn lots of threads/`compile` lots of networks in tests. Eitherway, seems to be something there, so glad we fixed this!

HeinrichApfelmus:
😱

Fortunately, I do not think that this is the indication of another bug. Rather, the `SomeNode` are stored in a `HashMap` and the most likely explanation is that you are seeing rare hash collisions where the `(==)` test is performed on nodes of different types.

PR #226, opened 2021-09-16
Generalise `compile` and `actuate` to `MonadIO`

ocharles:


mitchellwrosen (approved):
LGTM

vagarenko:
Perhabs lift `pause :: EventNetwork -> IO ()` too?

ocharles:
Good shout @vagarenko -  I think we should do all of these or none of them, so I'll add `pause` into this.

HeinrichApfelmus:
Alas, I'm not entirely sold on the introduction of MonadIO, as it makes the types more complicated.

I have made a comment in the related pull request https://github.com/HeinrichApfelmus/reactive-banana/pull/225#issuecomment-968275468 . To avoid duplication, I propose that we discuss this issue there.

PR #231, opened 2021-12-06
Add an `Eq (Ref a)` instance

ocharles:
This is necessary as the latest version of the `hashable` package has `Eq a => Hashable a`.

The `Eq` instance has been added to match the `Hashable` instance.


HeinrichApfelmus (approved):
Looks good, thank you! 😊

PR #232, opened 2021-12-17
Accept the majority of HLint suggestions

ocharles:


ocharles:
I'm getting back into some `reactive-banana` maintenance, and now that `haskell-language-server` has got better, a lot of things have come to light, as HLS enables HLint by default. I don't really care which hints we follow, but I would like to at least have a clean warning log. If you (@HeinrichApfelmus / @mitchellwrosen) could comment on any changes you disagree with, I will happily revert them and disable the HLint rule.

mitchellwrosen:
Cool, thanks for putting in the time to refactor and cleanup!

Most of these look fine to me. Some seem less readable, but that's just a personal preference.

A couple that stand out:

```haskell
-- before
(\m -> unM m 0)

-- after
(`unM` 0)
```

Is that better? I tend to stick to using infix notation for things that are clearly intended to be used infix, such as `Data.List.isPrefixOf`.

```haskell
-- before
(\time -> fmap (forgetE time)

-- after
(fmap . forgetE)
```

Pretty classic debate here but it seems weird for a tool to just authoritatively decide that all lambdas `\x -> f (g x)` should be written point-free. I think it should be up to us to decide whether e.g. naming the argument `time` improves readability.

```haskell
-- before
(\time xs -> xs !! time)

-- after
flip (!!)
```

Meh. I don't use `flip`, ever, for readability purposes. Again, just me.

```haskell
-- before
(\new old -> new ++ old)

-- after
(++)
```

Same point as above, but I thought I'd call out that the naming here was specifically requested in code review for readability purposes: https://github.com/HeinrichApfelmus/reactive-banana/pull/211#discussion_r625728571

```haskell
-- before
update (Lens get set) f = \s -> set (f $ get s) s

-- after
update (Lens get set) f s = set (f $ get s) s
```

Probably not an issue here, but this changes inlining behavior. A function can only inlined if applied to as many arguments appear on the LHS. So, something to be careful of when applying suggestions.

mitchellwrosen (approved):

ocharles:
@mitchellwrosen Thank you for the detailed review! I agree with all your points, and have updated my changes and included a `.hlint.yaml` that gets us back to zero HLint warnings.

HeinrichApfelmus:
Thanks, Charles and Mitchell! 😊 I agree with all of your points, Mitchell.

I actually find HLint to a be of a nuisance when used interactively, specifically for the issue of extraneous dollar signs `$`, because I find them very useful for structuring my code while writing it — often, I do not know in advance whether an argument will be a single identifier, or a full expression, and I can cover the oscillation between both variants by slapping a `$` in front of it. For example, the `$` in

     - mapL f lx = cachedLatch $ ({-# SCC mapL #-} f <$> getValueL lx)
     + mapL f lx = cachedLatch ({-# SCC mapL #-} f <$> getValueL lx)

was there first, but I had to add the parentheses in order to be able to attach the SCC pragma to the right expression.

But I do agree that removing `$` can be helpful to clean up the code just before a commit. I also like `fromMaybe`.

HeinrichApfelmus (approved):

ocharles:
> I actually find HLint to a be of a nuisance when used interactively, specifically for the issue of extraneous dollar signs `$`

Sometimes it can be useful to have separate HLint configurations - one for CI, and one for local development. Perhaps this is something we can explore in the future? Also, this is a good example of where a solid feedback loop is crucial. Now that HLint is integrated into haskell-language-server, one can have live feedback from HLint as they go. That, combined with the ability to automatically apply a hint, makes for a much nicer experience!

I'll go ahead and merge this as is now, and we can always tune the rules if they get annoying. I think it's also important to have this be part of a mandatory CI check, otherwise things gradually rot - but again, we can do this later.

HeinrichApfelmus:
> Sometimes it can be useful to have separate HLint configurations - one for CI, and one for local development. Perhaps this is something we can explore in the future?
> […]
> I'll go ahead and merge this as is now, and we can always tune the rules if they get annoying. I think it's also important to have this be part of a mandatory CI check, otherwise things gradually rot.

Agreed. Ok! 😊 Let's give it a try and change it later if interactive HLint becomes too cumbersome.

PR #233, opened 2021-12-17
Address warnings raised by `-Wall`

ocharles:


ocharles:
WIP, will tag once I'm happy with these changes!

ocharles:
We do have some build failures on 8.2 and below. I'm tempted to say we drop support for those, that's close to 5 years old and at this point almost no one uses it (according to https://taylor.fausak.me/2021/11/16/haskell-survey-results/#s2q4 and https://taylor.fausak.me/2020/11/22/haskell-survey-results/#s2q4)

evanrelf:
I recommend these too 🙂 

```
-Wcompat
-Werror=incomplete-record-updates
-Werror=incomplete-uni-patterns
-Werror=missing-fields
-Werror=partial-fields
```

mitchellwrosen:
> We do have some build failures on 8.2 and below. I'm tempted to say we drop support for those, that's close to 5 years old and at this point almost no one uses it (according to https://taylor.fausak.me/2021/11/16/haskell-survey-results/#s2q4 and https://taylor.fausak.me/2020/11/22/haskell-survey-results/#s2q4)

I'm also in favor of dropping support for compilers older than the 3 latest major versions, if it would ease our maintenance burden.

ocharles:
@evanrelf yes, those are all good! I'll add those in :+1:

HeinrichApfelmus:
I actually dislike `-Wname-shadowing`, because I sometimes use shadowing to ensure correctness. The prototypical example is a monadic sequence of state changes:

```
foo s = do
    s <- bar s
    s <- baz s
    pure s
```

Here, `s` denotes some intermediary state that is "changed" by `bar` and `baz`. By using shadowing, the question of "Which state do we use here?" always has a unique answer, because there is only a single `s` in scope! In comparison, in the code snippet
```
foo s0 = do
    s1 <- bar s0
    s2 <- baz s1
    pure s2
```
one has to be very careful which state is used when (`s0`, `s1` or `s2`?).

I have also moved away from using ticks as suffix to resolve shadowing, and prefer numbers instead, as numbers are easier to distinguish visually. I apply this to "state change" situation as above, but also to a situation like this

```
foo acc = go f acc
  where
    f acc = acc + 1
    …
```

where there is an initial value and a helper function describing a change. Instead of using ticks or underscores, I would now write this as

```
foo acc0 = go f acc0
  where
    f acc = acc + 1
    …
```

where the `0` makes it clear that `acc0` is the initial value, whereas `acc` is some intermediary value.

ocharles:
> I actually dislike -Wname-shadowing, because I sometimes use shadowing to ensure correctness. The prototypical example is a monadic sequence of state changes:

Thanks, I agree and was actually just about to revert that change - didn't think you'd replied :smile: What I wish we had in Haskell was a pragma to selectively say "I *want* this shadowing". Alas, such a thing doesn't exist.

> I have also moved away from using ticks as suffix to resolve shadowing, and prefer numbers instead, as numbers are easier to distinguish visually

I also usually do this - I learned I liked this after working on Dhall (https://hackage.haskell.org/package/dhall-1.40.2/docs/src/Dhall.Eval.html#normalize is a nice example).

Let me change this patch to `-Wall -Wno-name-shadowing` and lets see how we look then.

ocharles:
Ok! I pushed a new version and I think I'm happy with this. It's pretty conservative, and name shadowing hasn't been changed - I think it's being used to good affect at the moment, so a change here just makes things worse. But otherwise, I think this is a nice improvement. As it no longer builds on 8.2, I've also taken that out of CI.

reactive-banana/src/Reactive/Banana/Model.hs
```diff
@@ -2,6 +2,8 @@
     reactive-banana
 ------------------------------------------------------------------------------}
 {-# LANGUAGE RecursiveDo #-}
+{-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
+{-# OPTIONS_GHC -Wno-incomplete-patterns #-}
```
ocharles, line 6: I think the model really wants to be working with `data Stream a = a :> as`, but that's a change for another day. Rather than cluttering the model with exhaustive pattern matches that just use `error`, I've turned exhaustiveness checking off here.

mitchellwrosen (approved):

HeinrichApfelmus (approved):
Thanks a lot, Ollie! 😊

PR #234, opened 2021-12-18
Move `Reactive.Banana.Test.Performance to a proper `benchmark` suite

ocharles:
I'd like to start doing some performance investigations into `reactive-banana`, but the to do that, I need benchmarks! We do currently have a very basic benchmark in `Reactive.Banana.Test.Performance`, but there's no convenient way to actually run it. I've moved this benchmark out of `src` and into a new top-level `benchmark` folder, and added a `benchmark` component to `reactive-banana.cabal`. This new benchmark component uses the lightweight benchmarking library `tasty-bench` to run the benchmarks. A nice property of this library is that we could also actually use these benchmarks as regression tests, to make sure performance only gets better. My goal of this PR is mostly a refactoring though, so I'll leave that other work.

Here's what running this benchmark looks like:

```
❯ cabal run benchmark
Up to date
All
  netsize = 1
    duration = 1:   OK (0.24s)
      6.66 μs ± 479 ns
    duration = 2:   OK (0.12s)
      13.4 μs ± 1.3 μs
    duration = 4:   OK (0.22s)
      26.7 μs ± 1.8 μs
    duration = 8:   OK (0.22s)
      53.6 μs ± 3.0 μs
    duration = 16:  OK (0.22s)
      106  μs ± 5.7 μs
    duration = 32:  OK (0.11s)
      213  μs ±  21 μs
    duration = 64:  OK (0.22s)
      425  μs ±  30 μs
    duration = 128: OK (0.11s)
      858  μs ±  84 μs
  netsize = 2
    duration = 1:   OK (0.12s)
      7.00 μs ± 697 ns
    duration = 2:   OK (0.12s)
      13.9 μs ± 1.3 μs
    duration = 4:   OK (0.12s)
      27.7 μs ± 2.7 μs
    duration = 8:   OK (0.23s)
      55.1 μs ± 3.3 μs
    duration = 16:  OK (0.11s)
      110  μs ±  11 μs
    duration = 32:  OK (0.22s)
      220  μs ±  15 μs
    duration = 64:  OK (0.12s)
      442  μs ±  44 μs
    duration = 128: OK (0.23s)
      885  μs ±  43 μs
  netsize = 4
    duration = 1:   OK (0.24s)
      7.07 μs ± 351 ns
    duration = 2:   OK (0.12s)
      14.1 μs ± 1.4 μs
    duration = 4:   OK (0.23s)
      27.7 μs ± 1.4 μs
    duration = 8:   OK (0.12s)
[...]

PR #235, opened 2021-12-19
Move tests to `tests/` and switch test runner to `tasty`

ocharles:
There are two changes going on here.

1. The existing tests were previously located in `src`, but they are now in `tests`. The reason for this is organizational, and the current layout confuses some tooling, such as `haskell-language-server` (`hls`). By having everything in one directory, `hls` thinks that the test modules are actually part of the library component, and fails to build as not all dependencies are available in this component. Furthermore, having everything in `src` means that even when we build the test suite, we end up rebuilding everything in `src` - even though we just built it for the library. There is occasionally a good reason to do this, and you can test private modules. We don't do this, so as far as I can tell, there's no benefit having everything in one directory.
2. I've changed the test runner to `tasty`. I think this is slightly more modern/supported/active than `test-framework`, and as of #234 we already have a dependency on `tasty` via `tasty-bench`. It seems sensible to minimize our dependency graph and move to `tasty`.

### DIFF
--- a/reactive-banana/reactive-banana.cabal
+++ b/reactive-banana/reactive-banana.cabal
@@ -82,12 +82,12 @@ Library
 Test-Suite tests
     default-language:   Haskell98
     type:               exitcode-stdio-1.0
-    hs-source-dirs:     src
-    main-is:            Reactive/Banana/Test.hs
+    hs-source-dirs:     tests
+    main-is:            Main.hs
+    other-modules:      Plumbing
     build-depends:      base >= 4.2 && < 5,
-                        HUnit >= 1.2 && < 2,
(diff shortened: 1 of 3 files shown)


mitchellwrosen (approved):

ocharles:
I'm going to take @HeinrichApfelmus' :+1: as an approval and merge :smile:

HeinrichApfelmus:
👍👍👍 Thanks, Charles! 😊

I had thought that the tests did use internal modules, but that was probably from a time when multiple event occurrences per moment were allowed — it was easier to implement the "one occurrence per moment" version and then use a list for the occurrence on top of that. But those days are long gone.

PR #236, opened 2022-01-04
Build with `-O2`

ocharles:
The benchmarks speak for themselves here, I think!

First, with `-O0`:

```
$ cabal run benchmarks -O0
...
  netsize = 128
    ...
    duration = 128: OK (0.18s)
      13.1 ms ± 952 μs
```

Then, with `-O2`:

```
$ cabal run benchmarks -O2
...
  netsize = 128
    ...
    duration = 128: OK (0.20s)
      770  μs ±  47 μs
```

That is, the longest benchmark test goes from 13.1ms, down to 770μs - almost 20x faster.

### DIFF
--- a/reactive-banana/reactive-banana.cabal
+++ b/reactive-banana/reactive-banana.cabal
@@ -78,7 +78,7 @@ Library
                         Reactive.Banana.Prim.Util,
                         Reactive.Banana.Types
     
-    ghc-options: -Wall -Wcompat -Werror=incomplete-record-updates -Werror=incomplete-uni-patterns -Werror=missing-fields -Werror=partial-fields -Wno-name-shadowing
+    ghc-options: -O2 -Wall -Wcompat -Werror=incomplete-record-updates -Werror=incomplete-uni-patterns -Werror=missing-fields -Werror=partial-fields -Wno-name-shadowing
 
 Test-Suite tests
     default-language:   Haskell98


ocharles:
I haven't looked into *why* this is the case, but if I were to guess, it would be more aggressive inlining of `RWSTIO` and friends.

HeinrichApfelmus:
Wow! That looks like a worthwhile improvement.

> I haven't looked into why this is the case, but if I were to guess, it would be more aggressive inlining of RWSTIO and friends.

That would be my guess, too — but there is one possibility that we should exclude: That the benchmarks became faster because the optimizer has accidentally moved the time consuming stuff out of the measurement window. 😅

ocharles:
Yea, this is worth checking. I think if we build just the library with `-O2` we should be safe there. Note that they way these benchmarking frameworks work is it's meant to be really hard to accidentally so this kind of thing, but I'll double-check

mitchellwrosen:
@HeinrichApfelmus I think it's that.

I'm on mobile right now, but just after a cursory review of tasty-bench, the code looks very wrong.

I see INLINE pragmas on things that should absolutely be NOINLINE, but even that's not enough - for the core benchmark loops you must also prevent let-floating, either by tricking GHC with some fanciness, or just using -fno-full-laziness. 

(Shameless plug for my own thing) this is the correct way to write such benchmarking utils: https://github.com/awkward-squad/park-bench/blob/main/src/ParkBench/BenchmarkInternal.hs

mitchellwrosen:
@HeinrichApfelmus I think it's that.

I'm on mobile right now, but just after a cursory review of tasty-bench, the code looks very wrong.

I see INLINE pragmas on things that should absolutely be NOINLINE, but even that's not enough - for the core benchmark loops you must also prevent let-floating, either by tricking GHC with some fanciness, or just using -fno-full-laziness. 

(Shameless plug for my own thing) this is the correct way to write such benchmarking utils: https://github.com/awkward-squad/park-bench/blob/main/src/ParkBench/BenchmarkInternal.hs

mitchellwrosen:
Oh, but comparing `-O0` to `-O2`? Yeah, I'd except a huge difference.

I thought the benchmarks would run at `-O1` by default with `cabal run benchmarks`, and that would include building the library at `-O1`, too. 🤔

ocharles:
Ok, I had a look again. This time round, I'm just doing `$ cabal run benchmarks`. The `-O` stuff is set in `ghc-options` - I set `benchmarks` to build with `-O0`, and the `library` component to build with increasing levels of optimisations. Here are the results for `netsize = 128, duration = 128`:

* With `-O0`: 13.2 ms ± 849 μs
* With `-O1`: 1.11 ms ±  86 μs
* With `-O2`: 1.08 ms ±  85 μs

So it does seem that `-O2` isn't actually doing anything, and that the real gains were coming from optimizing the benchmarks themselves, as hypothesized.

HeinrichApfelmus:
Thanks Ollie! It was worth a try. 😄 And the difference between `-O0` and `-O1` *is* an order of magnitude.

Bodigrim:
> I'm on mobile right now, but just after a cursory review of tasty-bench, the code looks very wrong.
> https://github.com/Bodigrim/tasty-bench/blob/7ac455f14d2f93b5d49ff07e89503981aa249a77/Test/Tasty/Bench.hs#L1106
> I see INLINE pragmas on things that should absolutely be NOINLINE

I'm surprised with the abruptness of your evaluation: both [`gauge`](https://hackage.haskell.org/package/gauge-0.2.5/docs/src/Gauge.Benchmark.html#pureFunc) and [`criterion`](https://hackage.haskell.org/package/criterion-1.3.0.0/docs/src/Criterion-Types.html#nf) (prior to 1.4) INLINE these functions. This was changed (without much fuss) in `criterion-1.4` to NOINLINE, but I'm not quite convinced with the reasoning, since it affects inlining of `rnf`. Happy to continue discussion at `tasty-bench` issue tracker.

PR #237, opened 2022-01-07
Make sure that firing an `AddHandler` doesn't allocate

ocharles:
Edit: Below is the original version of the PR, where I changed the internal `Map` to be a list. We've since found a way to keep the use of `Map`, but to remove the need to allocate.

-------

Currently the creation of an `AddHandler` using `newAddHandler` uses a `Map` for bookkeeping, allowing for registrations to be unregistered. While this means lookup for deletion is fast, it incurs a penalty for registration and event dispatch. I'm primarily interested in the event dispatch point, as I think the main use of `AddHandler` is to fire them, whereas registration and unregistration is less frequent.

The use of a `Map` in event dispatch seems to incur an unavoidable allocation. If we dump out the current STG for `newAddHandler` (with `-O2`), we see:

```haskell
newAddHandler1 =
    \r [s_s4a3]
        case newMutVar# [Tip s_s4a3] of {
        (#,#) ipv_s4a5 ipv1_s4a6 ->
        let {
          sat_s4b0 =
              \r [a1_s4aG s1_s4aH]
                  case readMutVar# [ipv1_s4a6 s1_s4aH] of {
                  (#,#) ipv2_s4aJ ipv3_s4aK ->
                  let {
                    Rec {
                    go16_s4aL =
                        \r [z'_s4aM ds2_s4aN eta_s4aO]
                            case ds2_s4aN of {
                              Bin _ _ x_s4aS l_s4aT r_s4aU ->
                                  let {
                                    sat_s4aZ =
                                        \r [s2_s4aV]
                                            case x_s4aS a1_s4aG s2_s4aV of {
                                            (#,#) ipv4_s4aX _ -> go16_s4aL z'_s4aM r_s4aU ipv4_s4aX;
                                            };
                                  } in  go16_s4aL sat_s4aZ l_s4aT eta_s4aO;
                              Tip -> z'_s4aM eta_s4aO;
                            };
                    end Rec }
                  } in  go16_s4aL newAddHandler2 ipv3_s4aK ipv2_s4aJ;
                  }; } in
  ... let { sat_s4b1 = CCCS (,)! [sat_s4aF sat_s4b0];
        } in  (#,#) [ipv_s4a5 sat_s4b1];
```

`sat_s4b0` is the second element of the tuple returned from `newAddHandler` - this is the `a -> IO ()` function that fires the event. If we study `sat_s4b0`, we can see it allocates `go16_s4aL` whenever it's called (and even worse, this recursive function also allocates, though single-registration `AddHandler`s wouldn't be affected by that).

I can't see a way to fold a `Map Unique (a -> IO ())` into a single `IO ()` without allocating, so my suggestion is we instead change `Map Unique (a -> IO ())` into `[(Unique, a -> IO ())]`. If we do this, we can instead have STG like:

```haskell
newAddHandler1 =
    \r [s_s3SA]
        case newMutVar# [[] s_s3SA] of {
        (#,#) ipv_s3SC ipv1_s3SD ->
        let {
          sat_s3TB =
              \r [a1_s3Tk s1_s3Tl]
                  case readMutVar# [ipv1_s3SD s1_s3Tl] of {
                  (#,#) ipv2_s3Tn ipv3_s3To ->
                  let-no-escape {
                    Rec {
                    go1_s3Tp =
                        \r [ds2_s3Tq eta_s3Tr]
                            case ds2_s3Tq of {
                              [] -> (#,#) [eta_s3Tr ()];
                              : y_s3Tt ys_s3Tu ->
                                  case y_s3Tt of {
                                  (,) _ y1_s3Tx ->
                                  case y1_s3Tx a1_s3Tk eta_s3Tr of {
                                  (#,#) ipv4_s3Tz _ -> go1_s3Tp ys_s3Tu ipv4_s3Tz;
                                  };
                                  };
                            };
                    end Rec }
                  } in  go1_s3Tp ipv3_s3To ipv2_s3Tn;
                  }; } in
```

Here the "event firing" function uses `let-no-escape`, which doesn't allocate.

The impact of this is observable. I wrote a really dumb benchmark:

```haskell
    boringBenchmark = withResource setup mempty $ \getEnv ->
      bench "Boring" $ whnfIO $ do
        tick <- getEnv
        {-# SCC ticks #-} replicateM_ 10_000_000 $ {-# SCC tick #-} tick ()
      where
        setup = do
          (tick, onTick) <- newAddHandler
          compile $ do
            e <- fromAddHandler tick
            reactimate $ return <$> e
          return onTick
```

The change is:

* Baseline (`cabal run benchmarks -- +RTS -s`): 268  ms ±  26 ms, 1.2 GB allocated, 1.2 MB copied,  37 MB peak memory
* With this change: 202  ms ±  20 ms, 608 MB allocated, 1.2 MB copied,  37 MB peak memory


ocharles:
Of course, if anyone can figure out a way to `mapM_` over a `Map` without allocations, that might be an even better change!

ocharles:
Ok, thanks to a great suggestion from @istandleet on Twitter, I've found a way to do this using `Map`. The fix is to change `runHandlers` to:

```haskell
flip runReaderT a . getAp . foldMap (Ap . ReaderT) =<< readIORef handlers
```

A little gnarly, but if we do this, we get:

```
Rec {
newAddHandler2 =
    \r [ds_srKJ eta_srKK eta1_srKL]
        case ds_srKJ of {
          Bin dt_srKN _ v_srKP ds2_srKQ ds3_srKR ->
              case dt_srKN of {
                __DEFAULT ->
                    case newAddHandler2 ds2_srKQ eta_srKK eta1_srKL of {
                    (#,#) ipv_srKU _ ->
                    case v_srKP eta_srKK ipv_srKU of {
                    (#,#) ipv2_srKX _ ->
                    case newAddHandler2 ds3_srKR eta_srKK ipv2_srKX of {
                    (#,#) ipv4_srL0 _ -> (#,#) [ipv4_srL0 ()];
                    };
                    };
                    };
                1# -> v_srKP eta_srKK eta1_srKL;
              };
          Tip -> (#,#) [eta1_srKL ()];
        };
end Rec }

newAddHandler3 =
    \r [old_srL2]
        case integerAdd old_srL2 newUnique2 of _new1_srL3 {
        __DEFAULT -> (,) [_new1_srL3 _new1_srL3];
        };

newAddHandler1 =
    \r [s_srL4]
        case newMutVar# [Tip s_srL4] of {
        (#,#) ipv_srL6 ipv1_srL7 ->
        let {
          sat_srLM =
              \r [a1_srLH s1_srLI]
                  case readMutVar# [ipv1_srL7 s1_srLI] of {
                  (#,#) ipv2_srLK ipv3_srLL ->
                  newAddHandler2 ipv3_srLL a1_srLH ipv2_srLK;
                  }; } in
```

Which has no allocations.

So if we'd prefer to keep `Map`, I can update this PR accordingly. I might also add a `weight` test to confirm that there are no allocations.

HeinrichApfelmus:
Awesome, thanks Ollie! I'd prefer to keep the `Map`, indeed. 😊

Could you give the Twitter combination a name and a type signature, maybe like this?

```haskell
let runAllOn :: a → Map k (a → IO ()) → IO ()
    runAllOn a = flip runReaderT a . getAp . foldMap (Ap . ReaderT)
runAllOn a =<< readIORef handlers
```

In this way, we will have a chance to remember what this function does in a couple of years. 😄

mitchellwrosen (approved):
Woo, nice job!

ocharles:
@HeinrichApfelmus sure, all sounds good! I'll also add a comment linking to this PR so we know what's going on

HeinrichApfelmus (approved):

ocharles:
Ok, back to using `Map`, and i've double checked the benchmarks/STG. Thanks for the help, i'll merge this now!

ocharles:
One more snag: `Data.Monoid.Ap` is only in GHC 8.6 onwards. Are we OK dropping 8.4, or should I just manually define it? Or perhaps on `8.4` I can just use the sub-optimal `mapM_`?

HeinrichApfelmus:
I would be in favor of using CPP and conditionally adding the following definitions:

```haskell
newtype Ap f a = Ap { getAp :: f a }

instance (Applicative f, Semigroup a) => Semigroup (Ap f a) where
    (Ap x) <> (Ap y) = Ap $ liftA2 (<>) x y

instance (Applicative f, Monoid a) => Monoid (Ap f a) where
    mempty = Ap $ pure mempty
```

Maybe in an extra module `Data.Compat`? The idea being that this module collects functions that have made it into the `base` package already, just not for the given compiler version.

ocharles:
@HeinrichApfelmus How about depending on `base-compat` then, which is exactly that? https://hackage.haskell.org/package/base-compat-0.12.1/docs/Data-Monoid-Compat.html#t:Ap

HeinrichApfelmus:
> @HeinrichApfelmus How about depending on base-compat then, which is exactly that? https://hackage.haskell.org/package/base-compat-0.12.1/docs/Data-Monoid-Compat.html#t:Ap

Yes, please! 😊 I concur that this is the right way to go about here.

ocharles:
Doh, I was wrong `base-compat` explicitly avoids adding data types. I'll go with the CPP suggestion

PR #239, opened 2022-01-10
Add new `once` combinator

### LINKED ISSUE #74: Consider adding a `once` primitive

ocharles:
Fixes #74

### DIFF
--- a/reactive-banana/src/Reactive/Banana/Combinators.hs
+++ b/reactive-banana/src/Reactive/Banana/Combinators.hs
@@ -353,7 +353,7 @@ split e = (filterJust $ fromLeft <$> e, filterJust $ fromRight <$> e)
 -- | Keep only the next occurence of an event.
 --
 -- > once e = \time0 -> take 1 [(t, a) | (t, a) <- e, time0 <= t]
-once :: Event a -> MomentIO (Event a)
+once :: MonadMoment m => Event a -> m (Event a)
 once e = mdo
     e1 <- switchE e (never <$ e1)
     return e1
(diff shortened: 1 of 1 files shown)


mitchellwrosen (commented):
LGTM, but shouldn't it be in MonadMoment?

ocharles:
Oh yes, good catch!

mitchellwrosen (approved):

reactive-banana/src/Reactive/Banana/Combinators.hs
```diff
@@ -349,6 +350,15 @@ split e = (filterJust $ fromLeft <$> e, filterJust $ fromRight <$> e)
     fromRight (Right b) = Just b
 
 
+-- | Keep only the next occurence of an event.
+--
```
HeinrichApfelmus, line 354:
```suggestion
-- (Also aids the garbage collector by indicating that the result event can be discarded after its only occurrence.)
```

HeinrichApfelmus (approved):
Looks great, thanks! Could you commit my suggestion concerning garbage collection? The it's ready to go!

reactive-banana/src/Reactive/Banana/Combinators.hs
ocharles: Sounds good! I'm going to slightly reword this without parens, as it seems important enough to stand alone.

PR #240, opened 2022-01-10
Remove references to old test files.

Invisible-Rabbit-Hunter:
It seems that when @ocharles moved the test files to their own directory (#235), they forgot to remove references to those files in the extra-source-files field of the cabal file. This leads to an error when trying to use Reactive Banana as a `source-repository-package` in a cabal.project, as cabal while complain that such files do not exist.


ocharles:
Great catch, thank you!

Issue #228, opened 2021-09-28
Successful evaluation for an undefined network


ocharles:
I believe the following FRP system has undefined behavior:

```haskell
{-# language LambdaCase, BlockArguments,  RecursiveDo #-}
module Test where

import Reactive.Banana
import Reactive.Banana.Frameworks
import Reactive.Banana.Combinators
import Data.Functor

main :: IO ()
main = do
  (ah1, doStart) <- newAddHandler
  (ah2, doStep) <- newAddHandler

  actuate =<< compile mdo
    onStartManual <- fromAddHandler ah1
    onStep <- fromAddHandler ah2

    let onStart = unionWith const onStartManual ("Won't see me!" <$ filterJust (fst <$> onStateChange))

    onStateChange <- accumE (Nothing, []) $ unions
      [ onStart <&> \y (_, xs) -> (Nothing, xs ++ [y])
      , onStep $> \case
          (_, x:xs) -> (Just x, xs)
          (_, [])   -> (Nothing, [])
      ]

    reactimate $ putStrLn "onStartManual" <$ onStartManual
    reactimate $ putStrLn "onStart" <$ onStart
    reactimate $ print <$> onStateChange

  doStart "Hello"
  doStep ()
```

The problem is `onStart` is defined in terms of `onStateChange`, but `onStateChange` is defined in terms of `onStart`. This is mutual recursion between events, and the documentation writes

> In general, mutual recursion between several Events and Behaviors is always well-defined, as long as an Event depends on itself only via a Behavior, and vice versa.

That's not the case here, so this recursion isn't well defined. What concerns me is that it *does* run though! What we see in practice is:

```
> main
onStartManual
onStart
(Nothing,["Hello"])
onStart
(Just "Hello",[])
```

This is quite confusing:

1. `onStartManual` occurs, causing `onStart` to occur
2. `onStateChange` occurs, moving the state from `(Nothing, [])` to `(Nothing, ["Hello"])` - this is expected.
3. `onStep` occurs, causing `onStateChange` to change from `(Just "Hello", [])`.
4. However, at this instant we now know that `onStart` should occur, so the state transition should really have been the union of *both* `onStep` and `onStart`, so we'd expect to be at `(Just "Hello", ["Won't see me!"])`. This doesn't happen, and we instead enter a state that I think is impossible

I wonder if it's possible to spot this impossible recursion and to crash. I would much rather things crash than enter strange states.

[...]

PR #241, opened 2022-01-16
Add first draft of design document + rearrange modules

HeinrichApfelmus:
I have begun writing a draft on the design internals of the reactive-banana library. I had always wanted to write this up, as I believe that there are a couple of intricate design issues solved, but have never found the time to do so — until now.

Documenting the internals is useful for understanding and fixing issues such as #228 .

I have also rearranged the modules to follow the "design levels" (low, mid, high) in an attempt to make the edifice easier to understand. The idea is that understanding the implementation of a lower level is not required for understanding the implementation of a higher level.

Question to my dear co-maintainers: As writing will still be slow in the foreseeable future, which areas in the `design.md` document would you like me to expand on next? Anything that you are particularly interested in or find particularly puzzling?

### DIFF
--- a/reactive-banana/doc/design/design.md
+++ b/reactive-banana/doc/design/design.md
@@ -1,0 +1,77 @@
+# Design and Implementation of the reactive-banana FRP library
+
+by Heinrich Apfelmus
+
+licensed CC-BY-SA 4.0
+
+DRAFT, INCOMPLETE
+
+This document is intended to describe the design and implementation of the reactive-banana library for functional reactive programming (FRP).
+
+NOTE: This draft text already uses the plural `Events` as opposed to the singular `Event` for the data type.
+
+## Introduction
+
+Functional reactive programming (FRP) is a programming paradigm for declarative programming with data that changes over time. It was introduced in the seminal paper [Elliot and Hudak (1997)][fran] with two concepts, *Behavior* and *Events*. A *Behavior* denotes a value that varies in time, i.e. *Behavior* is a data type that represents the entire history and future of a value. *Events* denote a sequence of discrete moments in time which carry additional data.
+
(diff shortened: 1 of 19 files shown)


ocharles:
Oh, what a terrific idea! I've actually been thinking about setting up a [Read the Docs](https://readthedocs.io) page for `reactive-banana`, and a design document is one section I'd like to have there (along side more tutorials, a cookbook, etc). I was going to try and write the design document myself, but it's even better if it comes from the source!

> Anything that you are particularly interested in or find particularly puzzling?

I feel quite confident that I understand the internals of `reactive-banana` from the recent performance work I've been doing, and I think my understanding matches your low/mid/high stratification. My understanding is that at the most primitive level we have a graph of pulses. We enrich this with latches to introduce history, and then we further enrich this to get the lovely API that we give to consumers.

The missing piece of the puzzle to me is observable sharing. I know what it is (at least, in as much as I know [`data-reify`](https://hackage.haskell.org/package/data-reify) and have read *Type-Safe Observable Sharing in Haskell* (Gill 09)). What haven't quite formed an understanding of *how* this is used in `reactive-banana`, or why it is necessary. Expanding that section would be the highest value to me.

mitchellwrosen:
Looks great so far! I don't have any specific requests for areas to document next.

Speaking of observable sharing, there is a small thread about it in this issue, with an unresolved comment at the bottom: https://github.com/HeinrichApfelmus/reactive-banana/pull/221

It might be good to circle back at some point before documenting why `Cached` is the way it is.

HeinrichApfelmus:
> The missing piece of the puzzle to me is observable sharing [..]  Expanding that section would be the highest value to me.

Great, thanks! When I'll get around to tackling the next section, it will be the section on observable sharing. 😊 

> Speaking of observable sharing, there is a small thread about it in this issue, with an unresolved comment at the bottom: #221

Absolutely. I'll get back to that.

> There are unfortunately important operational concerns (or perhaps even abstraction leak?) to users. […]

Good points; these are mostly concurrency issues that need to be documented. Overall, I have opted to implement a design that does not introduce any concurrency on its own, but is stable under concurrent use.

PR #242, opened 2022-01-16
Make Network always contain the always-pulse

mitchellwrosen:
This PR replaces the

```haskell
nAlwaysP :: Maybe (Pulse ())
```

field of `Network` with

```haskell
nAlwaysP :: Pulse ()
```

and propagates that change around the codebase.

Previously, when compiling a network, we checked to see whether it had an always-pulse, and built it if not.

In this PR, we instead just create the always-pulse when initializing an empty network in IO.

### DIFF
--- a/reactive-banana/src/Reactive/Banana/Prim/Compile.hs
+++ b/reactive-banana/src/Reactive/Banana/Prim/Compile.hs
@@ -21,7 +21,7 @@ import           Reactive.Banana.Prim.Types
 -- | Change a 'Network' of pulses and latches by
 -- executing a 'BuildIO' action.
 compile :: BuildIO a -> Network -> IO (a, Network)
-compile m Network {nTime, nOutputs, nAlwaysP} = do
+compile m Network{nTime, nOutputs, nAlwaysP} = do
     (a, topology, os) <- runBuildIO (nTime, nAlwaysP) m
     doit topology
 
--- a/reactive-banana/src/Reactive/Banana/Prim/Compile.hs
+++ b/reactive-banana/src/Reactive/Banana/Prim/Compile.hs
@@ -2,6 +2,7 @@
     reactive-banana
 ------------------------------------------------------------------------------}
 {-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE NamedFieldPuns #-}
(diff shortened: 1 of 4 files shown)


ocharles (approved):
I'm in favor of this! It eliminates a `case` that occurs every time we step the network. This is minor, but a nice little win

reactive-banana/src/Reactive/Banana/Prim/Compile.hs
```diff
@@ -20,26 +20,26 @@ import           Reactive.Banana.Prim.Types
...
-        outputs1 = nOutputs state1
-
-    theAlwaysP <- case nAlwaysP state1 of
-        Just x   -> return x
-        Nothing  -> do
-            (x,_,_) <- runBuildIO undefined $ newPulse "alwaysP" (return $ Just ())
-            return x
-
-    (a, topology, os) <- runBuildIO (nTime state1, theAlwaysP) m
+compile m (Network time1 outputs1 theAlwaysP) = do
```
ocharles, line 23: I would recommend using record syntax (`NamedFieldPuns` would be my preference, too).

HeinrichApfelmus (changes requested):
Thanks Mitchell! 😊 What Ollie said, I'm in favor as well, with the little change.

reactive-banana/src/Reactive/Banana/Prim/Compile.hs
```diff
@@ -20,14 +21,14 @@ import           Reactive.Banana.Prim.Types
 -- | Change a 'Network' of pulses and latches by
 -- executing a 'BuildIO' action.
 compile :: BuildIO a -> Network -> IO (a, Network)
-compile m (Network time1 outputs1 theAlwaysP) = do
-    (a, topology, os) <- runBuildIO (time1, theAlwaysP) m
+compile m Network {nTime, nOutputs, nAlwaysP} = do
```
HeinrichApfelmus, line 24:
```suggestion
compile m Network{nTime, nOutputs, nAlwaysP} = do
```

It's probably a matter of taste, but I like to leave no space between the constructor and the curly braces, lest they can be confused for two separate arguments.

HeinrichApfelmus (approved):

PR #252, opened 2022-03-24
Documentation on observable sharing

HeinrichApfelmus:
This pull request adds an explanation of how reactive-banana implements observable sharing.

### DIFF
--- a/reactive-banana/doc/design/design.md
+++ b/reactive-banana/doc/design/design.md
@@ -35,7 +35,9 @@ For example, consider a graphical user interface (GUI) with a button and a numer
 
 The easiest way to implement Events would be as a list of values tagged with their time of occurrence, e.g.
 
-    type Events a = [(Time, a)]
+```hs
+type Events a = [(Time, a)]
+```
 
 By using a list lazy, future events which are not yet known can be added to the list when they occur.
 
@@ -51,7 +53,9 @@ The necessity of implementing `Events` as a data structure that supports updates
 
 2. *Mid-level*. We define low-level data structure `Pulse` and `Latch` that roughly correspond to Events and Behavior, but which are essentially just references to nodes in the mutable graph above. New nodes can be created with monadic functions such as
 
-        union :: Pluse a -> Pulse a -> Build (Pulse a)
+    ```hs
+    union :: Pluse a -> Pulse a -> Build (Pulse a)
+    ```
 
     where `Build` is a monad that describes changes to the mutable graph.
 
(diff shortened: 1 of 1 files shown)


reactive-banana/doc/design/design.md
```diff
@@ -56,22 +57,113 @@ The necessity of implementing `Events` as a data structure that supports updates
...
+This section explains how observable sharing is used in the implementation of `Events` and `Behavior`.
+
+*Observable sharing* is a technique that aids the implementation of *domain specific languages* (DSL) that are *embedded* in a purely functional host language. It makes the variable binding and sharing mechanism of the host language (e.g. Haskell's `let`) partially observable in the embedded language in order to avoid duplicating large subexpressions ([Claessen and Sands 1999][clsa], [Gill 2009][gill], [Kiselyov 2011][oleg],…). We use this technique for the same purpose in our implementation of FRP.
+
+To explain observable sharing, let us first consider a toy example and only then consider an FRP example. The toy example is a small language for evaluating arithmetic expressions `Exp`:
+
+    data Exp
+        = Val Int     -- plain `Int`
+        | Add Exp Exp -- addition
+        | Mul Exp Exp -- multiplication
```
ocharles, line 96: I'd suggest writing these code blocks as:

    ```haskell
    data Exp =  ...
    ```

That way we'll get syntax highlighting.

reactive-banana/doc/design/design.md
```diff
@@ -56,22 +57,113 @@ The necessity of implementing `Events` as a data structure that supports updates
...
+    x * x
+```
+where `var = pure . Var`. The code is almost identical, except that the "macro expansion" case `y` uses the host language `let` construct, while the "shared variable" case `y'` uses the monadic assignment `<-`, which essentially serves as a (variant of the) embedded language `Let`. Now, the purpose of *observable sharing* is to make it so that `y` generates the same expressions as `y'`, and the way to do that is to change the behavior of the "macro" `x :: M Name`: Instead of creating a new expression every time that this monadic action is executed, it creates an expression only once, stores the result in a variable `x'`, and returns `var x'` whenever it is executed (this can only be done using `unsafePerformIO`).
+
+Reactive-banana […]
+
+
+  [clsa]: http://www.cse.chalmers.se/~dave/papers/observable-sharing.pdf 
+  [gill]: http://www.ittc.ku.edu/~andygill/papers/reifyGraph.pdf
+  [oleg]: https://arxiv.org/abs/1109.0784s
```
Icelandjack, line 169: the trailing *s* doesn't belong in the link: https://arxiv.org/abs/1109.0784

HeinrichApfelmus:
I finally found some time to finish this up. 😊  @ocharles , does this explanation work for you, or do you want me to expand on a couple of points?

HeinrichApfelmus:
(Merging this for now, feel free to request clarifications or more content.)

Issue #152, opened 2017-07-25
Space / time leak


nickkuk:
The following code leaks for reactive-banana-1.1.0.1 and GHC-8.0.2:
```
module Main where

import Control.Monad
import Reactive.Banana.Frameworks
import Reactive.Banana

main :: IO ()
main = do
  (addHandler, fire) <- newAddHandler
  let nw = do
        e1 <- fromAddHandler addHandler
        e2 <- switchE (never <$ e1)
        reactimate $ return () <$ e2
  compile nw >>= actuate
  forever $ fire ()
```

ocharles:
Could you add an explicit GC after `fire ()` to confirm?

nickkuk:
Do you mean
```
module Main where

import Control.Monad
import System.Mem
import Reactive.Banana.Frameworks
import Reactive.Banana

main :: IO ()
main = do
  (addHandler, fire) <- newAddHandler
  let nw = do
        e1 <- fromAddHandler addHandler
        e2 <- switchE (never <$ e1)
        reactimate $ return () <$ e2
  compile nw >>= actuate
  forever $ do
    fire ()
    performMajorGC
```
? It seems that memory grows too, but much slower.

nickkuk:
@ocharles Or it stop to grow at some value... Does it exists robust way to confirm that there is no leak for long living program?

ocharles:
That's what I meant, I just wanted to verify it was a real leak, and not just GC not kicking in.

vagarenko:
@nickkuk 
There is a way
http://neilmitchell.blogspot.com/2015/09/detecting-space-leaks.html

nickkuk:
@vagarenko It seems that aforementioned code doesn't use any stack. I think that leaks in reactive-banana library (if they're exist) should be more like ones from the last point in "Caveats" section from [http://neilmitchell.blogspot.com/2015/09/detecting-space-leaks.html](http://neilmitchell.blogspot.com/2015/09/detecting-space-leaks.html), i.e. heap leaks.

nickkuk:
@ocharles @HeinrichApfelmus I can confirm that memory consumption grows with performMajorGC too. It starts with 3 MB and eats 200 MB during the night.

HeinrichApfelmus:
Om nom nom. That's unusual and shouldn't happen, I think. 🤔 Unfortunately, it will take a while until I can look into this, I'm currently a bit swamped.

DevJac:
[...]

PR #255, opened 2022-06-12
Fix space leak in dynamic event switching

### LINKED ISSUE #253: Unexpected memory growth
The following simplified code seems to leak memory. Manually running a major GC after every `fireTick ()` does not solve the problem.

I'm using 054781d0b0bf4e43ec1f395e3ccdbe7c7bd25896

```haskell
main :: IO ()
main = do
    (tickHandler, fireTick) <- newAddHandler
    let
        networkDescription :: MomentIO ()
        networkDescription = do
            tick <- fromAddHandler tickHandler
            x <- switchE tick (tick <$ tick)
            reactimate (print () <$ x)
            pure ()
    net <- compile networkDescription
    actuate net
    replicateM_ 50000 $ do
        fireTick ()
```

image

ocharles:
Currently we have a nasty space leak in dynamic event switching. The crux of the problem is that we have a child event that needs to have a new parent. When we change the parent of a node, we need to create a weak pointer to the new parent and store it in the child's list of parents, and we need to store a weak pointer to the child in the new parent node.

Currently `connectChild` does this by creating a new weak pointer to the child whenever the parent changes:

```haskell
w <- mkWeakNodeValue child child
```

Here we create a weak pointer to the child `SomeNode`, kept alive by the underlying `IORef` that backs `child`.

The problem with this approach is that the `child` `IORef` may be reachable throughout the entire duration of the program execution (e.g., `child` is connected to a top-level `reactimate`). This is problematic, because a weak pointer is kept alive by the garbage collector if the key of the weak pointer is alive *regardless* of whether anyone actually has a reference to the weak pointer itself. In the case of `switchE`, this means any time we switch a child to have a new parent, we allocate one more weak pointer that can never be garbage collected.

The fix is to instead create these self-referential weak pointers whenever pulses, latches and outputs are created. This means that when we switch parent, we can avoid allocating new weak pointers to the (stable) child node (essentially caching the weak pointer).

Fixes #253, #152.

### DIFF
--- a/reactive-banana/src/Reactive/Banana/Prim/Low/Dependencies.hs
+++ b/reactive-banana/src/Reactive/Banana/Prim/Low/Dependencies.hs
@@ -48,7 +48,7 @@ connectChild
     -> IO (Weak SomeNode)
                 -- ^ Weak reference with the child as key and the parent as value.
 connectChild parent child = do
-    w <- mkWeakNodeValue child child
+    w <- someNodeRef child
     modify' parent $ update childrenP (w:)
     mkWeakNodeValue child (P parent)        -- child keeps parent alive
 
--- a/reactive-banana/src/Reactive/Banana/Prim/Low/IO.hs
+++ b/reactive-banana/src/Reactive/Banana/Prim/Low/IO.hs
@@ -28,15 +28,19 @@ newInput :: forall a. Build (Pulse a, a -> Step)
 newInput = mdo
     always <- alwaysP
     key    <- liftIO Lazy.newKey
-    pulse  <- liftIO $ newRef $ Pulse
-        { _keyP      = key
-        , _seenP     = agesAgo
-        , _evalP     = readPulseP pulse    -- get its own value
-        , _childrenP = []
-        , _parentsP  = []
-        , _levelP    = ground
-        , _nameP     = "newInput"
-        }
+    pulse  <- liftIO $ mdo
+        me <- newRef $ Pulse
+            { _keyP      = key
+            , _seenP     = agesAgo
+            , _evalP     = readPulseP pulse    -- get its own value
+            , _childrenP = []
+            , _parentsP  = []
+            , _levelP    = ground
+            , _nameP     = "newInput"
+            , _pulsePtr  = w
+            }
+        w <- mkWeakRefValue me (P me)
+        return me
     -- Also add the  alwaysP  pulse to the inputs.
     let run :: a -> Step
         run a = step ([P pulse, P always], Lazy.insert key (Just a) Lazy.empty)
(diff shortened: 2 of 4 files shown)


mitchellwrosen (approved):
LGTM, nice work! The explanation makes sense to me and the code looks good. Just curious, have you confirmed manually that the space leak is gone? Or, should we try to write a test that demonstrates `switchE` doesn't leak?

Coincidentally - and off-topicly - I have a little wrapper around System.Mem.Weak that I was considering publishing soon. You might find it interesting/cute (or not, because you clearly already understand how to wield regular old mkWeak): https://hackage.haskell.org/package/ersatz-pointer-0.1.0/candidate/docs/ErsatzPointer.html

ocharles:
> Just curious, have you confirmed manually that the space leak is gone? Or, should we try to write a test that demonstrates switchE doesn't leak?

Yep, I ran the examples in both linked issues and with this commit both examples run in constant memory. A test would be good, I can see if `weigh` is useful here.

I'm a bit more puzzled why the line two lines down (which also creates a weak pointer with the child as a key) _doesnt_ leak, but I can investigate that separately. I have feeling that this weak pointer is getting finalized, which is why this pointer doesn't leak.


> Coincidentally - and off-topicly - I have a little wrapper around System.Mem.Weak that I was considering publishing soon. You might find it interesting/cute (or not, because you clearly already understand how to wield regular old mkWeak): https://hackage.haskell.org/package/ersatz-pointer-0.1.0/candidate/docs/ErsatzPointer.html

Very cool! I understand the dark arts of weak pointers *now*, but it took almost two full days at ZuriHac! Maybe you want to try and adapt reactive-banana to use this interface - I'd certainly be interested to see it (though not certain if I'd merge it)

mitchellwrosen:
> Yep, I ran the examples in both linked issues and with this commit both examples run in constant memory. A test would be good, I can see if weigh is useful here.

Excellent, I think that's sufficient. No real desire from me to add an automated test, unless you want to.

> I'm a bit more puzzled why the line two lines down (which also creates a weak pointer with the child as a key) doesnt leak, but I can investigate that separately. I have feeling that this weak pointer is getting finalized, which is why this pointer doesn't leak.

Hmm... yeah. What's up with that? The only way this weak pointer would get finalized is if the child gets GC'd, but I thought it wasn't, which was the cause of the original leak. I think I'm missing something here -- a diagram would be awful useful :)

> Maybe you want to try and adapt reactive-banana to use this interface - I'd certainly be interested to see it (though not certain if I'd merge it)

Maybe! But I kinda think it's only really useful as an alternative explanation of weak pointer concepts, probably not worth an extra dependency for library code. However, I think privately refactoring reactive-banana to use that interface is a good idea anyway, to feel out whether the API is actually better :)

ocharles:
> Hmm... yeah. What's up with that? The only way this weak pointer would get finalized is if the child gets GC'd, but I thought it wasn't, which was the cause of the original leak. I think I'm missing something here -- a diagram would be awful useful :)

I worked it out on the flight back from ZuriHac! The reason this one leaks is because there is never a corresponding call to `finalize` (which I actually learnt about by reading `ersatz-pointer`'s source!). This presents an alternative fix: we can just add a call to `finalize` in `isGoodChild` if we've found ourselves in an old parent. I need to have a think about which is better, and I'll publish a PR with my alternative version shortly.

mitchellwrosen:
Oh, great :tada:

PR #256, opened 2022-06-13
Fix space leak in dynamic event switching

### LINKED ISSUE #253: Unexpected memory growth

ocharles:
Currently we have a nasty space leak in dynamic event switching. The crux of the problem is that we have a child event that needs to have a new parent. When we change the parent of a node, we need to create a weak pointer to the new parent and store it in the child's list of parents, and we need to store a weak pointer to the child in the new parent node.

Currently `connectChild` does this by creating a new weak pointer to the child whenever the parent changes:

```haskell
w <- mkWeakNodeValue child child
```

Here we create a weak pointer to the child `SomeNode`, kept alive by the underlying `IORef` that backs `child`.

The problem with this approach is that the `child` `IORef` may be reachable throughout the entire duration of the program execution (e.g., `child` is connected to a top-level `reactimate`). This is problematic, because a weak pointer is kept alive by the garbage collector if the key of the weak pointer is alive *regardless* of whether anyone actually has a reference to the weak pointer itself. In the case of `switchE`, this means any time we switch a child to have a new parent, we allocate one more weak pointer that can never be garbage collected.

The fix to allow the old weak pointers to be garbage collected is to call `finalize` on them, which we now do in `removeParents`. `removeParents` is responsible for removing a child from all current parents, so when we remove ourself from a parent (removing the weak pointer from the parent), we finalize the weak pointer.

Fixes #253, #152.

### DIFF
--- a/reactive-banana/src/Reactive/Banana/Prim/Low/Dependencies.hs
+++ b/reactive-banana/src/Reactive/Banana/Prim/Low/Dependencies.hs
@@ -76,6 +76,12 @@ removeParents child = do
               case x of
                 Just y | y /= P child -> return True
                 _                     -> do
+                  -- The old parent refers to this child. In this case we'll remove
+                  -- this child from the parent, but we also need to finalize the
+                  -- weak pointer that points to the child. We need to do this because
+                  -- otherwise the weak pointer will stay alive (even though it's
+                  -- unreachable) for as long as the child is alive
+                  -- https://github.com/HeinrichApfelmus/reactive-banana/pull/256
                   finalize w
                   return False
         new <- filterM isGoodChild . _childrenP =<< readRef parent
--- a/reactive-banana/src/Reactive/Banana/Prim/Low/Dependencies.hs
+++ b/reactive-banana/src/Reactive/Banana/Prim/Low/Dependencies.hs
@@ -72,7 +72,12 @@ removeParents child = do
     forM_ _parentsP $ \w -> do
         Just (P parent) <- deRefWeak w  -- get parent node
         finalize w                      -- severe connection in garbage collector
-        let isGoodChild w = not . maybe True (== P child) <$> deRefWeak w
+        let isGoodChild w = deRefWeak w >>= \x ->
+              case x of
+                Just y | y /= P child -> return True
+                _                     -> do
+                  finalize w
+                  return False
         new <- filterM isGoodChild . _childrenP =<< readRef parent
         modify' parent $ set childrenP new
     -- replace parents by empty list


ocharles:
I think I prefer this approach as we don't have to maintain an invariant in all `Pulse`s/`LatchWrite`s/`Output`s, which in the other PR have to store a weak pointer to themselves. There's a minor cost to this PR though, in that it means whenever we change parents we allocate and call `finalize`, which does increase the cost of dynamic switching. I'm curious which approach others prefer.

HeinrichApfelmus:
> I think I prefer this approach

I prefer this approach as well, as finalizing the weak pointer is very much in line with the actions that you would naturally expect `removeParents` to do.

The most curious thing is probably that calling `finalize` on a weak pointer will make the weak pointer unreachable — this behavior is *not* documented in `System.Mem.Weak`! The documentation only says "A heap object is reachable if … It is a weak pointer object whose key is reachable." We may want to consider a small patch to the GHC documentation. 🤔

ocharles:
Indeed - it took three of us two days at ZuriHac to figure this out! It is documented, but in true Haskell fashion - only via a paper :facepalm: See [Stretching the Storage Manager](https://www.microsoft.com/en-us/research/wp-content/uploads/1999/09/stretching.pdf), section 5.3 and again in 5.5. I agree docs could be better though!

HeinrichApfelmus:
> I agree docs could be better though!

Perhaps you could add a small comment to that effect in the code — e.g. mention that `finalize w` does make `w` unreachable? For posterity and eternal fame.

HeinrichApfelmus (approved):

HeinrichApfelmus:
> It is documented, but in true Haskell fashion - only via a paper 🤦 See [Stretching the Storage Manager](https://www.microsoft.com/en-us/research/wp-content/uploads/1999/09/stretching.pdf), section 5.3 and again in 5.5.

When originally writing the code, I had spent a lot of time with this paper — and was a bit disappointed when GHCJS implemented something less sophisticated. 😅

ocharles:
Added a comment with a link back to this PR

HeinrichApfelmus (approved):

HeinrichApfelmus:
Thanks! Feel free to merge pull requests that @HeinrichApfelmus has ✅ approved with a Github review. 🤓🙏

PR #258, opened 2022-06-14
Force new `Network` after completing a step

ocharles:
Minor, but currently when we step the event network we produce a new `Network`, but nothing forces this. `step` completes and returns a thunk containing the network, which gets stored in an `MVar`. This thunk isn't adding anything, so instead it would be better to just force the `Network`. We could do this either in `step` itself (as done in this commit), or we could force it when we store it the `s` `MVar`.


mitchellwrosen:
(I'm on mobile, otherwise I would look around myself) Does Network have strict fields? If not, I think we need more seqs n' bangs than just that one. :)

ocharles:
Yep, it does indeed have strict fields already.

mitchellwrosen (approved):

PR #260, opened 2022-06-14
Implement `accumL` with `newPulse`

ocharles:
Currently `accumL` uses `mapL` to read the previous state whenever the state update `Pulse` fires. This is problematic, because when the previous state is read in `mapL` (and transformed into a `(a -> a) -> a` function), this is cached into the `Latch`. This means that when the state update `Pulse` fires, we'll update one latch to the new state, while also retaining the previous state. This commit redefines `accumL` as a call to `newLatch` (as before) and creates a new `Pulse` with `newPulse` whose evaluation function reads the `Pulse` containing the `a -> a` endmorphism, and also reads the current state from its latch. With this implementation, no historic values are leaked by `accumL`.

### DIFF
--- a/reactive-banana/src/Reactive/Banana/Prim/Mid/Combinators.hs
+++ b/reactive-banana/src/Reactive/Banana/Prim/Mid/Combinators.hs
@@ -98,7 +98,11 @@ applyL lf lx = cachedLatch
 accumL :: a -> Pulse (a -> a) -> Build (Latch a, Pulse a)
 accumL a p1 = do
     (updateOn, x) <- newLatch a
-    p2 <- applyP (mapL (\x f -> f x) x) p1
+    p2 <- newPulse "accumL" $ do
+      a <- readLatchP x
+      f <- readPulseP p1
+      return $ fmap (\g -> g a) f
+    p2 `dependOn` p1
     updateOn p2
     return (x,p2)
 


mitchellwrosen:
I'm not 100% clear on what's going on but I trust that this makes things better. At the very least I can see how this allocates one fewer latch nodes, so that seems good.

> This means that when the state update Pulse fires, we'll update one latch to the new state, while also retaining the previous state.

A few questions - mostly permutations of the same question, really - where are we retaining the previous state? Do you mean in memory somewhere, due to laziness or something? (Like, does the `(\x f -> f x)` thing matter here, or is this transformation something we should do to all `applyP (mapL ...)` things?) I'd have naively guessed that the latch is updated to the latest state via its latch write node, which is a child of pulse we make, so I'm not seeing where the historical values are leaked. Is it a space leak that grows and grows, or just a single-step leak?

mitchellwrosen (approved):
I'm

HeinrichApfelmus:
Thanks for catching this and sorry for taking so long to take a look at this -- work. 🤪

I agree with you, @ocharles

* the 'mapL' creates an additional Latch, whose value is going to be cached
* moreover, this value has a function type, and those can often contain more data than necessary

I approve, feel free to merge. 🙏🏻

HeinrichApfelmus (approved):

ocharles:
@HeinrichApfelmus You need a job like mine where `reactive-banana` is part of your work :wink: Thanks for the review, no worries about taking your time - I'll get this merged now!

mitchellwrosen:
Whoa, who's hiring banana engineers?!

ocharles:
CircuitHub :sunglasses: But not hiring at the moment unfortunately :disappointed:

HeinrichApfelmus:
> CircuitHub 😎 But not hiring at the moment unfortunately 😞

If they feel the need to hire again — you know where to find us. 😉

PR #267, opened 2022-12-25
Implement a graph with garbage collection

HeinrichApfelmus:
### Overview

This pull request implements a type `GraphGC` that represents a graph whose vertices are subject to garbage collection.

Eventually, we will use `GraphGC` to represent a network of `Pulse`: The vertices are `Pulse`, and the edges between them are dependencies. The function `walkSuccessors` allows us to walk the `Pulse` in dependency order — with early exit (`Stop`), but still with the guarantee that no vertex will be visited before its dependencies.

### Comments

* Garbage collection is tricky, hence we use a pure `Graph` to tests the `GraphGC` against (In fact, we use it for most of the implementation as well, except for the actual garbage collection). We implement an ordinary function `collectGarbage` on the pure graph and use QuickCheck to test that real garbage collection produces the same results.
* Garbage collection will mark garbage for collection, but not remove it outright. In order to remove edges from the `GraphGC`, one has to call `removeGarbage` explicitly — this makes it easier to reason about which vertices are in the graph, because we no longer have to deal with the concurrency of garbage collection.
* Using the pure `Graph` data type in the implementation of `GraphGC` also has the advantage that the correctness of `walkSuccessors` can be checked on the pure data type.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -89,14 +89,16 @@ jobs:
       - name: Build dependencies
         run: cabal build reactive-banana:lib:reactive-banana --disable-optimization --only-dependencies --with-compiler ghc-${{matrix.ghc}}
 
+      # We need to build the library and tests with the default optimization level
+      # in order to test garbage collection
       - name: Build library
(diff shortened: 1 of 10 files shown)


HeinrichApfelmus:
… and the failing CI indicates that testing garbage collection with QuickCheck is harder than I thought. It worked locally! :joy:

Issue #261, opened 2022-06-15
Memory leak with dynamic behavior switching


ocharles:
The following program:

```haskell
{-# language BlockArguments #-}
module Main where

import Control.Monad
import Data.Functor
import Reactive.Banana
import Reactive.Banana.Frameworks
import System.Mem
import System.Mem.Weak

withGhcDebug = id

main :: IO ()
main = withGhcDebug do
  (ah1, fire1) <- newAddHandler

  actuate =<< compile do
    e <- fromAddHandler ah1

    let e2 = observeE $ e $> do
          stepper () (void e)

    b' <- switchB (pure ()) e2

    reactimate $ return <$> b' <@ e

  performGC
  putStrLn "Running"

  replicateM_ 50000 $ do
    fire1 ()
    performGC
```

Leaks memory:

![image](https://user-images.githubusercontent.com/20878/173808610-c4b69fbf-78aa-4370-b13a-89a3b82ffd2d.png)

I've also modified `doAddChild` to print out the number of children in a parent:

```haskell
doAddChild :: SomeNode -> SomeNode -> IO ()
doAddChild (P parent) (P child) = do
    level1 <- _levelP <$> readRef child
    level2 <- _levelP <$> readRef parent
    let level = level1 `max` (level2 + 1)
    w <- parent `connectChild` P child
    print =<< length . _childrenP <$> readRef parent -- <- NEW
    modify' child $ set levelP level . update parentsP (w:)
doAddChild (P parent) node = void $ parent `connectChild` node
doAddChild (L _) _ = error "doAddChild: Cannot add children to LatchWrite"
doAddChild (O _) _ = error "doAddChild: Cannot add children to Output"
```

This shows that a node has a 50000 children at the end. My guess is this is the `Event` `e` - whenever `execute` fires we attach a new stepper to `e`, but this child is never removed - even though `switchB` should be discarding them.

mitchellwrosen:
Nice, can you share how you made the pretty graph?

ocharles:
Sure! I just used [eventlog2html](https://hackage.haskell.org/package/eventlog2html) :smile: 

More specifically:

```
$ cabal run leak -- +RTS -l -hi
```

and build with `-rtsopts -eventlog`.

I often also build with `-finfo-table-map -fdistinct-constructor-tables`, as per https://well-typed.com/blog/2021/01/first-look-at-hi-profiling-mode/

HeinrichApfelmus:
[...]

PR #268, opened 2022-12-26
Perform evaluation steps `Network` using `GraphGC`

HeinrichApfelmus:
### Overview

This pull request completely changes the way that dependencies between `Pulse` are tracked and used in order to perform an `Evaluation.step` for a `Network`.

We use the machinery provided by `GraphGC` to

* track dependencies between `Pulse`, using `insertEdge` and `clearPredecessors`.
* traverse the `Pulse` in dependency order with early exit, using `walkSuccessors_`.

### Comments

* This should fix many remaining issues with garbage collection for `Pulse`, specifically #261 
* I think that in order to fix *all* remaining issues for `Pulse`, we may have to look at garbage collection and `Vault`.
* This pull request doesn't do anything for `Latch`.

Still, 🥳!

### Obsoletes

* #182
* sadly, #243


HeinrichApfelmus:
Alas, it appears that the event log for Ollie's example of https://github.com/HeinrichApfelmus/reactive-banana/issues/261#issuecomment-1279988614 still shows the presence of a space leak. 🤔 Needs more work.

Eventually, the right thing to do is to include a few tests in the test suite that can detect certain space leaks. 🤓 In other words — instead of checking for space leaks manually, we should automate checking for them.

That said, the new tests do show that garbage collection on `GraphGC` works in principle, so I'm not sure what's going on with Ollie's example. 🤔 Trying to automate the `-evenlog` GHC flag could be tricky, so I'll probably start with pure measure of the `GraphGC` size.

HeinrichApfelmus:
> Alas, it appears that the event log for Ollie's example of https://github.com/HeinrichApfelmus/reactive-banana/issues/261#issuecomment-1279988614 still shows the presence of a space leak.

Actually, setting `profiling: True` in `cabal.project` and running

```
$ cabal run leak -- +RTS -hT -RTS
$ hp2pretty leak.hp
```

results in the following heap profile:

leak


The program completes very quickly and runs in constant size. 🤔 

However! Compiling the program with `profiling: False` will increase the running time considerably — there seems to be a space leak.

HeinrichApfelmus:
> Eventually, the right thing to do is to include a few tests in the test suite that can detect certain space leaks. 🤓 In other words — instead of checking for space leaks manually, we should automate checking for them.

I have added a function `getSize` that measures the size of the event network and used it to turn the offending program #261 into a test case — and voilà, it appears that the leak is caused by more and more pulses being added to the network.

Twist: The test will *pass*  when I set `profiling: True` — but another test, `observeE_stepper` will fail! It looks like there is something wrong in the (new) interplay between `Pulse` and `Latch`.

HeinrichApfelmus:
Getting closer.

I have added a test case that works on the `Mid` level as opposed to the `High` level; the test checks whether the combination of `executeP` and `accumL` adds more and more pulses to the graph — and this is currently the case. This is fortunate, as this result rules out observable sharing as root cause for the leak.

Observation: If I remove `performSufficientGC` from the `runNetworkSizes` function, then the network will grow *faster*. This implies that some garbage collection is actually taking place, it's just not fast enough at removing old pulses, or it retains some parts for longer than necessary.

HeinrichApfelmus:
Finally, fixed it! 🥳

Turns out that I was accidentally relying on `tryReadTQueue` to read all vertices scheduled for deletion from the queue — but that will only read a single item. The function that I actually wanted is `flushTQueue`.

HeinrichApfelmus:
Hm, the space profile for #261 is still not constant, but the leak is not in the network graph, at least. 🤔

leak

HeinrichApfelmus:
I found the culprit for the remaining space leak: In the function `deleteVertex`, I forgot to delete the vertex from the `levels` map in `Graph`:

https://github.com/HeinrichApfelmus/reactive-banana/blob/4aa23eabf80ea737719c4481764aba80c8403e16/reactive-banana/src/Reactive/Banana/Prim/Low/Graph.hs#L179-L180

Here is the space profile after fixing this:

leak

🥳

PR #273, opened 2023-02-24
add `activate` as the preferred naming of `actuate`

mitchellwrosen:
This PR renames `actuate` to `activate`, as approved in #147 some 6 years ago ;)

I've left `actuate` around (of course) without a DEPRECATED pragma, as I feel this would probably be annoying to users. But, I figure we can add one eventually, if we really want to delete the symbol from the interface some day.


dpwiz:
What got you change your mind?

HeinrichApfelmus (approved):
6 years have passed and I feel none the wiser. Approved, thanks!

PR #275, opened 2023-02-24
small refactor: make BuildW a proper data type

mitchellwrosen:
This PR just gives field names to each of the four components of a `BuildW`.

This would be a breaking change requiring a major version bump, except `BuildW` is conspicuously missing from the export list of [Reactive.Banana.Prim.Mid](https://hackage.haskell.org/package/reactive-banana-1.3.2.0/docs/Reactive-Banana-Prim-Mid.html) - unless we say that module doesn't follow the PVP - but was not exporting `BuildW` intentional?

### DIFF
--- a/reactive-banana/src/Reactive/Banana/Prim/Mid/Plumbing.hs
+++ b/reactive-banana/src/Reactive/Banana/Prim/Mid/Plumbing.hs
@@ -129,7 +129,7 @@ addOutput p = do
         { _evalO = fromMaybe (pure $ pure ()) <$> readPulseP p
         }
     _nodeP p `addChild` o
-    RW.tell mempty { bwOutputs = [o] }
+    RW.tell emptyBuildW{ bwOutputs = [o] }
 
 {-----------------------------------------------------------------------------
     Build monad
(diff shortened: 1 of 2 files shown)


reactive-banana/src/Reactive/Banana/Prim/Mid/Plumbing.hs
```diff
@@ -187,17 +187,17 @@ keepAlive child parent = liftIO $ void $
...
 changeParent pulse0 parent0 =
-    RW.tell $ BuildW ([ChangeParentTo pulse parent], mempty, mempty, mempty)
+    RW.tell mempty { bwDependencyChanges = [ChangeParentTo pulse parent] }
    where
      pulse = _nodeP pulse0
      parent = _nodeP parent0
 
 liftIOLater :: IO () -> Build ()
-liftIOLater x = RW.tell $ BuildW (mempty, mempty, Action x, mempty)
+liftIOLater x = RW.tell mempty { bwLateIO = Action x }
```
HeinrichApfelmus, line 200:
```suggestion
liftIOLater x = RW.tell emptyBuildW{ bwLateIO = Action x }
```

Two requests on the matter of record syntax:

* `mempty` is too generic a name. Could you define `emptyBuildW = mempty`, so that it better resembles the `BuildW` constructor?
* Record syntax messes a little with the "spaces separate distinct arguments" rule. Could you remove the spaces between the value and `{`? In this way, it becomes visually clear that `{` modifies the value.

HeinrichApfelmus:
`BuildW` is supposed to be an internal type — I'm happy to see it changed without bumping the version number. I consider [Reactive.Banana.Prim.Mid](https://hackage.haskell.org/package/reactive-banana-1.3.2.0/docs/Reactive-Banana-Prim-Mid.html) to be an internal module not subject to the PVP also.

reactive-banana/src/Reactive/Banana/Prim/Mid/Plumbing.hs
mitchellwrosen: You got it!

mitchellwrosen:
@HeinrichApfelmus I think this is good to go

HeinrichApfelmus:
Thanks! 😊

PR #276, opened 2023-02-24
refactor: delete Action newtype

mitchellwrosen:
A tiny PR with negligible worth but I figured I would put it up anyway ;)

This PR deletes the Action newtype in favor of IO's own semigroup instance, which has been around since GHC 8.2 (released in 2017).

Conflicts with #275


HeinrichApfelmus (approved):
Similar to `DependencyChanges`, I may eventually want to use an algebraic data type here ("defunctionalization"), but sure, why not. 🙂

PR #279, opened 2023-02-25
optimize a couple graph functions

mitchellwrosen:
This PR deletes a couple unnecessary O(n) `<$ ()`) calls. I suspect these were added at a time when graph `outgoing` and `incoming` had different types, but now they're the same.


HeinrichApfelmus (approved):
Good catch — thanks!

PR #280, opened 2023-02-25
delete seenP

mitchellwrosen:
It seems with the new internals we no longer rely on a timestamp to know if a pulse has been evaluated. Does that sound right? If so, this PR just deletes the unused field.


HeinrichApfelmus:
I think you're right. 🤔 Good catch, than you! Feel free to merge.

HeinrichApfelmus (approved):

PR #281, opened 2023-02-25
NOINLINE pureL

### LINKED ISSUE #37: Potential Bug: usage of unsafePerformIO?
In `Reactive.Banana.Internal.PulseLatch0` you are using `unsafePerformIO` several times without using a `NOINLINE` pragma. Is this intentional or could problems as described in [the documentation](http://www.haskell.org/ghc/docs/7.0-latest/html/libraries/base-4.3.1.0/System-IO-Unsafe.html) arise?

mitchellwrosen:
This pragma isn't completely necessary, but feels better than not having it. We can close #37 now.


HeinrichApfelmus (approved):
👍

PR #288, opened 2023-09-08
Allow pqueue-1.5

Bodigrim:


mitchellwrosen:
Hi, thanks, any other info? Have you looked at the changeling of the new version and confirmed the bump is safe, or is this PR just to get the conversation started?

Bodigrim:
I did, but you don't have to rely on my word only: the CI is green.

mitchellwrosen:
For sure. The type checker isn't always able to detect incompatibilities, so for bumps like this it helps to know a human has poked around the changelogs. Ok, thanks for the contribution!

Bodigrim:
@mitchellwrosen could you possibly please make a Hackage revision?

mitchellwrosen:
@Bodigrim done 👍

PR #290, opened 2023-11-15
slightly defunctionalize Latch by giving it a PureL constructor

mitchellwrosen:
This PR slightly defunctionalizes `Latch` by distinguishing (in the constructor tag) whether a latch is pure or not.

I gather from https://github.com/HeinrichApfelmus/reactive-banana/issues/37#issuecomment-1445208119 this is not necessarily the be-all end-all representation of Latch.

Nonetheless, it was easy to do, and I figured it might push the codebase forward a little bit. Feel free to close if not, though.

(Oh, some unrelated changes I see in the diff: I added an export list to `Mid.Combinators`)


HeinrichApfelmus:
Seems worth it to me. Thanks! 😊

PR #291, opened 2023-11-15
slightly improve performance of getting graph size

mitchellwrosen:
Just something small I noticed: we got the graph size by getting the size of `union(outgoing, incoming)`, but it's also available via the size of `levels`.

### DIFF
--- a/reactive-banana/src/Reactive/Banana/Prim/Low/Graph.hs
+++ b/reactive-banana/src/Reactive/Banana/Prim/Low/Graph.hs
@@ -124,11 +124,11 @@ listConnectedVertices :: (Eq v, Hashable v) => Graph v e -> [v]
 listConnectedVertices Graph{incoming,outgoing} =
     Map.keys $ outgoing `Map.union` incoming
 
--- | Number of connected vertices,
+-- | O(v) Number of connected vertices,
 -- i.e. vertices on which at least one edge is incident.
 size :: (Eq v, Hashable v) => Graph v e -> Int
-size Graph{incoming,outgoing} =
-    Map.size $ outgoing `Map.union` incoming
+size Graph{levels} =
+    Map.size levels
 
 -- | Number of edges.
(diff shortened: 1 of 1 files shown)


mitchellwrosen:
Hrm, sorry, this doesn't seem correct - I'm now noticing combinators like

```haskell
-- | Remove all the edges that connect the given vertex to its predecessors.
clearPredecessors :: (Eq v, Hashable v) => v -> Graph v e -> Graph v e
clearPredecessors x g@Graph{..} = g
    { outgoing = foldr ($) outgoing
        [ Map.adjust (Map.delete x) z | (z,_) <- getIncoming g x ]
    , incoming = Map.delete x incoming
    }
```

that don't keep `outgoing`/`incoming` in sync with `levels`.

There seem to be only two: `clearPredecessors` and `collectGarbage`. My gut instinct is that those functions ought to keep `levels` in sync with `outgoing`/`incoming`, or at least document why it's okay and intentional not to.

HeinrichApfelmus:
> Just something small I noticed: we got the graph size by getting the size of union(outgoing, incoming), but it's also available via the size of levels.

🤔 I think that this observation is correct in principle. In other words, I would agree that the invariant

```
Map.keys (outgoing `Map.union` incoming) == Map.keys levels
```

should hold. If you proceed with this, could you add this invariant as a comment to the `Graph` type?

* On the matter of `clearPredecessors`: I think it's correct for it to not adjust the `levels` — because only edges are removed, but not vertices. I.e. the invariant above should hold.

* On the matter of `collectGarbage`: The documentation says

    > -- | Apply `deleteVertex` to all vertices which are not predecessors
    > -- of any of the vertices in the given list.

    Now, `deleteVertex` does adjust the levels by calling `clearLevels `, but `collectGarbage` fails to do that. I think that's a mistake on the part of `collectGarbage`, it should restrict the keys of the `levels` Map to `reachables` as well.

PR #296, opened 2024-01-20
Allow reactive-banana-1.3

thielema:
I can compile reactive-banana-wx with latest reactive-banana.
Fortunately, wxhaskell:master/HEAD can built again with recent GHC versions and Nix.


HeinrichApfelmus:
Thanks! I have made a revision on Hackage.

> Fortunately, wxhaskell:master/HEAD can built again with recent GHC versions and Nix.

🥳

PR #298, opened 2024-12-22
Bump dependencies to allow `hashable-1.5.0.0`

HeinrichApfelmus:
This pull request updates the CI (with earlier cache of dependencies, but dropping the build cache) and bumps dependencies.


HeinrichApfelmus:
Testing locally shows that `hashable-1.5.0.0` compiles as well. I'm adding a metadata revision on Hackage.

